PR #4, opened 2016-10-17
langserver: optimize handling of vendor dependencies for workspace/reference

slimsag:
Previously we did a lot of work (typechecking and finding all external refs)
for vendor dependencies even though we would just throw out the results. By
not doing this work, we can transform the time it takes to invoke `workspace/reference`
for repos with many vendor dependencies drastically (e.g. prometheus goes
from ~5 mins -> 27s).

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -83,6 +83,22 @@ func (h *LangHandler) externalRefsFromPkg(ctx context.Context, bctx *build.Conte
 		return nil
 	}
 
+	if strings.HasPrefix(buildPkg.Dir, "vendor/") || strings.Contains(buildPkg.Dir, "/vendor/") {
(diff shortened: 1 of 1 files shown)


beyang:
Quinn asked me to review. This LGTM, but can we add 2 small test cases to test for the exclusion of vendored symbols from external references and the inclusion of external refs that aren't vendored?

slimsag:
@beyang there is no functional change, in fact this is behavior is already very well tested in langserver_test.go

slimsag:
more than one tests it, so check out this code + search for `wantWorkspaceReferences`) https://github.com/sourcegraph/sourcegraph-go/blob/master/langserver/langserver_test.go#L35-L426

I'm gonna merge, but if you want I can point out specific examples etc. just lmk

PR #5, opened 2016-10-18
vscode: add view external references

uforic:
Beta version of "View External References" button on VSCode. Video of using it in app.go on sourcegraph/sourcegraph: https://cl.ly/0u12322A3x2J

Works on all publicly available references, not references in private repos.

@sqs : Two issues I wanted to highlight:
- Private repositories don't work, because right now we call `https://sourcegraph.com/.api/repos/.../-hover-info` to construct the eventual def URL, but we can't access that URL unless we are auth'd. Ideally, directly from the LSP return we could construct the external references URL directly. I.E. /external-refs?file=<blah>&line=<blah>&col=<blah> , and it could just redirect to the right thing. We could also cut out a round trip here too.

(I recognize that Find External Refs doesn't really make sense in the case of private code... but perhaps someone at a large org wants to find external refs from other private repos in the company?) 
- Some weirdness with go standard library. I'm getting a file returned in /usr/local/go , but this is actually a homebrew repo. I special cased it, which is ugly, but so goes it:
  `
  // Special case homebrew
  if (gitUrl === "github.com/Homebrew/homebrew") {
      gitUrl = "github.com/golang/go";
      gitRev = GitUtils.getContentsOfFile(`${path.join(topLevelGitDirectory, "go")}/VERSION`);
      filePath = filePath.substr(3);
  }
  `
- The vendored repo case doesn't work - is there a way to get a rev of the repo? We don't need it I guess, I could add a special case to just point to master. Thoughts?


keegancsmith:
@uforic what is the status of this PR?

uforic:
Closing out! was an experiment to see how easy this would be.

Takeaway: URL construction would be really easy if the langserver was aware of things like vendoring, and able to provide some metadata around a given file.

i.e. vendor/github.com/gorilla/mux/mux.go , to construct the Sourcegraph URL, we need to read vendor.json, find the repository name (for github.com, can always assume foo/bar, but for gopkg.in, sometimes it's one word sometimes it's two. vendor.json lists the full name of the repo.

Also, this would be easier once the Microsoft language server has the functionality referenced here:
https://github.com/sourcegraph/language-server-protocol/pull/2 , so that we can avoid making two round trips to get the information required (parent, symbolname, etc.).

PR #9, opened 2016-10-20
Struct hover + workspace/reference output corrections

slimsag:
See individual commit messages for details.

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -44,7 +44,15 @@ func (h *LangHandler) handleHover(ctx context.Context, conn JSONRPC2Conn, req *j
 		// TODO(sqs): make this be like (T).F not "struct field F string".
 		s = "struct " + o.String()
 	} else if o != nil {
-		s = types.ObjectString(o, qf)
+		if obj, ok := o.(*types.TypeName); ok {
+			typ := obj.Type().Underlying()
+			if _, ok := typ.(*types.Struct); ok {
+				s = "type " + obj.Name() + " struct"
+			}
+		}
+		if s == "" {
+			s = types.ObjectString(o, qf)
+		}
 	} else if t != nil {
 		s = types.TypeString(t, qf)
 	}
--- a/langserver/internal/refs/refs.go
+++ b/langserver/internal/refs/refs.go
@@ -17,7 +17,10 @@ type Def struct {
 	// ImportPath is the import path at which the definition is located.
 	ImportPath string
 
-	// Path is the path of the definition (e.g. ["Router", "Route"] for a
+	// PackageName is the name the package is imported under.
+	PackageName string
+
+	// Path is the path of the definition (e.g. "Router Route" for a
 	// method named "Route" with receiver type "Router").
 	Path string
 }
@@ -32,7 +35,7 @@ type Ref struct {
 }
 
 func (r *Ref) String() string {
-	return fmt.Sprintf("&Ref{Def.ImportPath: %q, Def.Path: %q, Position: \"%s:%d:%d:%d\"}", r.Def.ImportPath, r.Def.Path, r.Position.Filename, r.Position.Line, r.Position.Column, r.Position.Offset)
+	return fmt.Sprintf("&Ref{Def.ImportPath: %q, Def.PackageName: %q, Def.Path: %q, Position: \"%s:%d:%d:%d\"}", r.Def.ImportPath, r.Def.PackageName, r.Def.Path, r.Position.Filename, r.Position.Line, r.Position.Column, r.Position.Offset)
 }
 
 type Config struct {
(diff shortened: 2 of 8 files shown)


langserver/langserver_test.go
```diff
@@ -83,7 +83,8 @@ func TestServer(t *testing.T) {
 				"a.go": "package p; type T struct { F string }",
 			},
 			wantHover: map[string]string{
-			// "a.go:1:28": "(T).F string", // TODO(sqs): see golang/hover.go; this is the output we want
+				// "a.go:1:28": "(T).F string", // TODO(sqs): see golang/hover.go; this is the output we want
```
keegancsmith: is this comment still relevant?

langserver/hover.go
```diff
@@ -44,7 +44,15 @@ func (h *LangHandler) handleHover(ctx context.Context, conn JSONRPC2Conn, req *j
 		// TODO(sqs): make this be like (T).F not "struct field F string".
 		s = "struct " + o.String()
 	} else if o != nil {
-		s = types.ObjectString(o, qf)
+		if obj, ok := o.(*types.TypeName); ok {
+			typ := obj.Type().Underlying()
+			if _, ok := typ.(*types.Struct); ok {
+				s = "type " + obj.Name() + " struct"
+			}
```
keegancsmith: I wonder if you can just do
```
s = fmt.Sprintf("type %s %s", obj.Name(), obj.Type().Underlying())
```
since `obj.Type().Underlying()` is a `types.Type` interface which requires `String() string`.

langserver/internal/refs/refs.go
```diff
@@ -255,6 +262,16 @@ func (c *Config) defInfo(pkg *types.Package, info *types.Info, rootFile *ast.Fil
...
 }
 
+// packageName returns name.Name if name != nil. Otherwise, it returns the last
+// element in importPath (e.g. "http" for "net/http")
+func packageName(importPath string, name *ast.Ident) string {
+	if name != nil {
+		return name.Name
+	}
+	split := strings.Split(importPath, "/")
+	return split[len(split)-1]
```
keegancsmith: not sure if this is optimising too soon, but you can avoid extra allocations by using `strings.LastIndex`.

langserver/internal/refs/refs_test.go
```diff
@@ -71,49 +71,49 @@ func TestParseFile(t *testing.T) {
...
 			},
 		},
 		{
 			Filename: "testdata/vars.go",
 			Want: []*Ref{
-				&Ref{Def: Def{ImportPath: "net/http", Path: ""}, Position: pos("testdata/vars.go:3:8 (offset 21)")},
-				&Ref{Def: Def{ImportPath: "net/http", Path: "Client"}, Position: pos("testdata/vars.go:6:14 (offset 74)")},
-				&Ref{Def: Def{ImportPath: "net/http", Path: "RoundTripper"}, Position: pos("testdata/vars.go:8:12 (offset 124)")},
-				&Ref{Def: Def{ImportPath: "net/http", Path: "Client Transport"}, Position: pos("testdata/vars.go:12:3 (offset 225)")},
-				&Ref{Def: Def{ImportPath: "net/http", Path: "Client"}, Position: pos("testdata/vars.go:11:12 (offset 194)")},
```
keegancsmith: the updated test cases don't really imply why this is important (later commits might though). From what I can tell this can just be computed from `ImportPath`, and if not, we should add a test for that.

langserver/langserver_test.go
slimsag: Yep, it is. This is still something we want to do, just not done via this PR.

langserver/hover.go
slimsag: I tried it just now, and the answer is no. The actual problem is that `obj.Type().Underlying().String()` is actually what returns the 'concrete type' i.e. `struct{...all fields...}` instead of just `struct` which reads a lot better.

```
--- FAIL: TestServer (0.04s)
    --- FAIL: TestServer/go_detailed (0.00s)
        --- FAIL: TestServer/go_detailed/hover-a.go:1:17 (0.00s)
        	langserver_test.go:560: got "type T struct{F string}", want "type T struct"
```

langserver/internal/refs/refs.go
slimsag: That def. would be optimizing too soon, and probably for something that doesn't matter, BUT that reads a lot cleaner IMO so done :)

langserver/internal/refs/refs_test.go
slimsag: Good point. I will add a test case for it. It's because there can be both `package foo` and `package foo_test` in the same directory.

slimsag: Also, you encouraging me to write a test for this also just made me uncover a bug that makes:

```
// packageName returns name.Name if name != nil. Otherwise, it returns the last
// element in importPath (e.g. "http" for "net/http")
func packageName(importPath string, name *ast.Ident) string {
	if name != nil {
		return name.Name
	}
	return importPath[strings.LastIndex(importPath, "/")+1:]
}
```

completely unneeded / incorrect :)

slimsag:
Addressed all feedback + added `go test -i ./...` to solve a (very confusing) CI problem 😄 (thanks @shurcooL for helping me solve it!)

Merging.

PR #10, opened 2016-10-24
langserver: add documentation strings to textDocument/hover

slimsag:
This makes it such that hovering over any symbol always provides the
relevant documentation for that symbol.

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -1,18 +1,21 @@
 package langserver
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"go/ast"
 	"go/token"
 	"go/types"
+	"strings"
 
+	"github.com/slimsag/godocmd"
 	"github.com/sourcegraph/jsonrpc2"
 	"github.com/sourcegraph/sourcegraph-go/pkg/lsp"
 )
 
 func (h *LangHandler) handleHover(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) (*lsp.Hover, error) {
-	fset, node, pkg, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
+	fset, node, prog, pkg, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
 	if err != nil {
 		// Invalid nodes means we tried to click on something which is
 		// not an ident (eg comment/string/etc). Return no information.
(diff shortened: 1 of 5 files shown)


langserver/hover.go
```diff
@@ -73,3 +106,43 @@ func packageStatementName(fset *token.FileSet, files []*ast.File, node *ast.Iden
...
+func maybeAddComments(comments string, contents []lsp.MarkedString) []lsp.MarkedString {
+	if comments != "" {
+		var b bytes.Buffer
+		doc.ToHTML(&b, comments, nil)
+		return append(contents, lsp.MarkedString{
+			Language: "text/html",
+			Value:    b.String(),
+		})
+	}
+	return contents
```
shurcooL: Did you write it this way in order to be more consistent with the 2 funcs below?

Instead of:

```Go
if comments == "" {
    return contents
}
... (less indent)
```

langserver/langserver_test.go
```diff
@@ -425,6 +424,67 @@ func yza() {}
...
+// Header is like an HTTP header, only better.
+type Header struct {
+	// F is a string, too.
+	F string
+}
+`,
+			},
+			wantHover: map[string]string{
+				"a.go:7:9": "package p; Package p is a package with lots of great things. \n\n",
+				//"a.go:9:9": "", TODO: handle hovering on import statements (ast.BasicLit)
```
shurcooL: Is this a TODO for another PR, or forgotten about (since I have no way of knowing other than to ask)?

langserver/hover.go
```diff
@@ -73,3 +106,42 @@ func packageStatementName(fset *token.FileSet, files []*ast.File, node *ast.Iden
...
+	var b bytes.Buffer
+	doc.ToMarkdown(&b, comments, nil)
+	return append(contents, lsp.MarkedString{
+		Language: "markdown",
+		Value:    b.String(),
+	})
+}
+
+// packageDoc finds the documentation for the named package from it's files or
+// a superset of them.
```
shurcooL: I don't think this matches 1:1 to how real godoc extracts package documentation, but it's close.

It's probably fair to leave further fine-tuning and tweaking until later.

Minor, typo in that comment, it should be "from its files", not "from it is files".

Finally, I don't quite understand what you mean by "or a superset of them"?

slimsag: - Will fix typo.
- superset: Any `[]*ast.File` that is inclusive of the entire packages files, but may include more (e.g. files outside of the package).

shurcooL: I'd suggest clarifying the superset phrasing a bit. It's kinda hard to read what "them" refers to.

How about:

```Go
// packageDoc finds the documentation for the named package from its files or
// additional files.
```

slimsag:
thanks :)

PR #12, opened 2016-10-27
Implement textDocument/documentSymbol

slimsag:
`textDocument/documentSymbol` is a part of the LSP (https://github.com/sourcegraph/language-server-protocol/blob/master/protocol.md#document-symbols-request).

In particular, it is useful to us because of its similarity to `workspace/symbol` except restricted to a single file ('document'). This means that e.g. when redirecting from a code file's position -> def landing page we only have to ask for a single file's symbols in order to find the symbol's `Name, ContainerName` pair which identify it instead of every symbol in the workspace (in our case, repository). It's worth adding because, for large workspaces, it can change the time taken from ~15s -> 1s or so to perform this operation.

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -10,7 +10,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -31,27 +30,32 @@ type query struct {
 }
 
 // parseQuery parses a user's raw query string and returns a
-// structured representation of the query.
-func parseQuery(q string) query {
-	var qu query
-	toks := tokenizer.Split(strings.ToLower(q), -1)
-	index := 0
-	for _, tok := range toks {
-		if tok == "" {
+// structured representation of the query. Possible queries are:
+func parseQuery(q string) (qu query) {
+	// All queries are case insensitive.
+	q = strings.ToLower(q)
+
+	// Split the query into space-delimited fields.
+	for _, field := range strings.Fields(q) {
+		// Check if the field is a filter like `is:exported`.
+		if field == "is:exported" {
+			qu.filter = filterExported
 			continue
 		}
-		index++
-		if kind, isKeyword := keywords[tok]; isKeyword {
-			qu.kind = kind
-			continue
+
+		// Each field is split into tokens, delimited by periods or slashes.
+		tokens := strings.FieldsFunc(field, func(c rune) bool {
+			return c == '.' || c == '/'
+		})
+		for _, tok := range tokens {
+			if kind, isKeyword := keywords[tok]; isKeyword {
+				qu.kind = kind
+				continue
+			}
+			qu.tokens = append(qu.tokens, tok)
 		}
-		if filter, isFilter := parseFilter(tok); isFilter {
-			qu.filter = filter
-			continue
-		}
-		qu.tokens = append(qu.tokens, tok)
 	}
-	return qu
+	return
 }
 
 type filterType string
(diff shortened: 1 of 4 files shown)


uforic:
Sick!! thanks for doing this dude.

langserver/symbol.go
```diff
@@ -27,31 +27,37 @@ import (
...
 // parseQuery parses a user's raw query string and returns a
-// structured representation of the query.
-func parseQuery(q string) query {
-	var qu query
-	toks := tokenizer.Split(strings.ToLower(q), -1)
-	index := 0
-	for _, tok := range toks {
-		if tok == "" {
+// structured representation of the query. Possible queries are:
+func parseQuery(q string) (qu query) {
```
shurcooL: > Possible queries are:

Did you mean to include something more there?

shurcooL: Minor style issue, I would advise against changing the signature to return `(qu query)` over `query`. Just a do a `var qu query` inside.

See https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters.

However, this is an unexported func and you won't ever see it in godoc, so perhaps it's fine to keep as is. Up to you. I'm okay with it as is because it's unexported.

langserver/symbol.go
```diff
@@ -27,31 +27,37 @@ import (
...
+			if kind, isKeyword := keywords[tok]; isKeyword {
+				qu.kind = kind
+				continue
+			}
+			qu.tokens = append(qu.tokens, tok)
 		}
-		qu.tokens = append(qu.tokens, tok)
 	}
-	return qu
+	return
```
shurcooL: Explicit `return qu` is more readable, I would recommend you keep it.

langserver/symbol.go
```diff
@@ -87,9 +78,6 @@ var keywords = map[string]lsp.SymbolKind{
 	"const":   lsp.SKConstant,
 }
 
-// tokenizer is a regexp for tokenizing a raw user query string.
-var tokenizer = regexp.MustCompile(`[\.\s\/]+`)
```
shurcooL: ❤️

langserver/symbol.go
```diff
@@ -9,8 +9,8 @@ import (
 	"go/token"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
-	"regexp"
```
shurcooL: 😻

langserver/symbol.go
```diff
@@ -220,10 +212,21 @@ func toSym(name, container string, kind lsp.SymbolKind, fs *token.FileSet, pos t
 	}
 }
 
+// handleTextDocumentSymbol handles `textDocument/documentSymbol` requests for
+// the Go language server.
+func (h *LangHandler) handleTextDocumentSymbol(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lsp.DocumentSymbolParams) ([]lsp.SymbolInformation, error) {
+	f := strings.TrimPrefix(params.TextDocument.URI, "file://")
+	return h.handleSymbol(ctx, conn, req, query{file: f}, 0)
```
beyang: This could probably be made more efficient by only collecting the symbols from the file in question, instead of collecting all symbols and then filtering by file (`handleSymbol` iterates over each package and calls `getPackageSyms` for every package regardless of whether it contains the desired file).

langserver/symbol.go
slimsag: Meant to remove this. Not possible to easily list all possible queries, better to just read the code directly.

langserver/symbol.go
slimsag: I think you meant `collectFromPkg` not `getPackageSyms`.

You're right that `handleSymbol` here iterates over each package invoking `collectFromPkg`, but check out the modification to `collectFromPkg` on lines 295 - 300: it aborts early if the package does not contain the desired file.

It's true that we could _probably_ reduce this further to only operating on a single file (instead of all files in the package the file of interest resides in), but I imagine this would be a lot more involved for not a lot of immediate benefits.

langserver/symbol.go
slimsag: Per discussion on Slack, will leave it as-is. I also agree it's better left in the func signature when the func is unexported.

slimsag:
Thanks for the review! 😃

PR #14, opened 2016-10-30
langserver: support multiple gopaths

uforic:
@shurcooL was testing out the go-langserver, and noticed that his multiple GOPATH environment was triggering an error when the server was initialized.

Question for the reviewer - is there a more "correct" way to check if the go file in question contains a given GOPATH than string contains?

### DIFF
--- a/langserver/build_context.go
+++ b/langserver/build_context.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/sourcegraph/ctxvfs"
@@ -72,6 +73,15 @@ func fsBuildContext(ctx context.Context, orig *build.Context, fs ctxvfs.FileSyst
 	return ctxt
 }
 
+// From: https://github.com/golang/tools/blob/b814a3b030588c115189743d7da79bce8b549ce1/go/buildutil/util.go#L84
+// dirHasPrefix tests whether the directory dir begins with prefix.
+func dirHasPrefix(dir, prefix string) bool {
+	if runtime.GOOS != "windows" {
+		return strings.HasPrefix(dir, prefix)
+	}
+	return len(dir) >= len(prefix) && strings.EqualFold(dir[:len(prefix)], prefix)
+}
+
 // ContainingPackage returns the package that contains the given
 // filename. It is like buildutil.ContainingPackage, except that:
 //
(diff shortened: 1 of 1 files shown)


langserver/build_context.go
```diff
@@ -94,7 +97,13 @@ func ContainingPackage(bctx *build.Context, filename string) (*build.Package, er
 	if PathHasPrefix(filename, bctx.GOROOT) {
 		srcDir = bctx.GOROOT // if workspace is Go stdlib
 	} else {
-		srcDir = bctx.GOPATH
+		srcDir = gopaths[0] // by default, choose first gopath
+		for _, gopath := range gopaths {
+			if strings.Contains(pkgDir, gopath) {
```
sqs: See the `dirHasPrefix` bit at https://sourcegraph.com/github.com/golang/tools/-/blob/go/buildutil/util.go#L56:6 for the bulletproof implementation.

langserver/build_context.go
```diff
@@ -82,8 +82,11 @@ func fsBuildContext(ctx context.Context, orig *build.Context, fs ctxvfs.FileSyst
 // * if the file is in the xtest package (package p_test not package p),
 //   it returns build.Package only representing that xtest package
 func ContainingPackage(bctx *build.Context, filename string) (*build.Package, error) {
-	if !filepath.IsAbs(bctx.GOPATH) || strings.Contains(bctx.GOPATH, string(os.PathListSeparator)) {
-		return nil, fmt.Errorf("build context GOPATH must contain exactly 1 entry and it must be an absolute path (GOPATH=%q)", bctx.GOPATH)
+	gopaths := strings.Split(bctx.GOPATH, string(os.PathListSeparator))
```
shurcooL: There's no need to resort to string manipulation to do this task. There's a function in Go standard library for parsing GOPATH.

https://godoc.org/path/filepath#SplitList

shurcooL: There's also https://godoc.org/go/build#Context.SrcDirs which is slightly higher level. It omits non-existing directories, and includes GOROOT.

langserver/build_context.go
```diff
@@ -94,7 +107,13 @@ func ContainingPackage(bctx *build.Context, filename string) (*build.Package, er
 	if PathHasPrefix(filename, bctx.GOROOT) {
 		srcDir = bctx.GOROOT // if workspace is Go stdlib
 	} else {
-		srcDir = bctx.GOPATH
+		srcDir = "" // with no GOPATH, only stdlib will work
```
shurcooL: Should the default value (if pkgDir is not a prefix of any gopath workspaces) be `bctx.GOROOT` or empty string?

On line 95/108 above, it says:

```Go
srcDir = bctx.GOROOT // if workspace is Go stdlib
```

So the empty string here seems unexpected, and I just wanted to confirm this is intentional.

I also see that the next line does `srcDir = path.Join(srcDir, "src") + "/"`, which I think would produce unexpected results if `srcDir` is empty string.

shurcooL:
I've tested this branch and I can confirm it fixes the issue on my setup with > 1 GOPATH workspace, everything works properly with it.

See one inline comment about a potential issue.

langserver/build_context.go
uforic: Yo! I checked the current behavior if the langserver-go has no GOPATH, and it ends up trying to join the empty string to src/, so I made sure we are failing equally ungracefully.

I think alternatively, we could just return an error here if len(gopaths)==0, "You are trying to use the go langserver without a GOPATH on a non standard library" message. I don't know enough about the surrounding and subsequent calls to know if failing here would be preferred.

PR #15, opened 2016-11-02
optimize textDocument/symbol and workspace/symbol

slimsag:
- langserver: optimize textDocument/symbol method

This optimizes the textDocument/symbol method, in particular for one-off
operations. Prior to this change we would only omit a packages results
after it was imported (via `collectFromPkg`). This was silly, because we
can do the same operation on the package list created by `buildutil.ExpandPatterns`
before even starting a goroutine for `collectFromPkg`. This significantly
reduces the amount of round-tripping needed in terms of VFS operations
to import every Go package in the workspace. Now only the Go package specified
by the file is imported.

- langserver: add "dir:<some_dir>" query filter to workspace/symbol

This makes it such that anyone with knowledge of which directory the
symbols are located in (relative to the workspace root) can effectively
ask the language server to filter results only to that directory.
Primarily, this gives a substantial performance gain for large workspaces.


langserver/util.go
```diff
@@ -22,3 +22,7 @@ func PathTrimPrefix(s, prefix string) string {
 	}
 	return strings.TrimPrefix(s, prefix)
 }
+
+func PathEqual(a, b string) bool {
```
shurcooL: Does this need to be exported?

Since it's exported, it should be documented per [our style guide](https://github.com/sourcegraph/sourcegraph/blob/master/docs/style.md#sourcegraph-style-guide) which simply defers to https://github.com/golang/go/wiki/CodeReviewComments#doc-comments on this.

slimsag: Good point, doesn't need to be exported. Will update.

langserver/symbol.go
```diff
@@ -244,6 +254,21 @@ func (h *LangHandler) handleSymbol(ctx context.Context, conn JSONRPC2Conn, req *
 		par := parallel.NewRun(8)
 		pkgs := buildutil.ExpandPatterns(bctx, []string{pkgPat})
 		for pkg := range pkgs {
+			// If we're restricting results to a single file or dir, ensure the
+			// package dir matches to avoid doing unnecessary work.
+			if results.query.file != "" {
+				filePkgPath := path.Dir(results.query.file)
+				filePkgPath = PathTrimPrefix(filePkgPath, h.init.BuildContext.GOROOT)
```
sqs: seems incorrect technically, like if it contained multiple of those prefixes strangely like `/usr/local/go/src/usr/local/go/src/src/gopath/src/github.com/foo/bar` it would still result in `github.com/foo/bar` although it should fail.

slimsag:
@shurcooL @sqs updated

langserver/symbol.go
slimsag: Fair enough -- I've added an if-else for this case just now.

sqs:
@slimsag Would you expect this to work on Sourcegraph?

```
$ curl -sS --data '[{"id":0,"method":"initialize","params":{"rootPath":"git://github.com/kubernetes/kubernetes?6184ebdcb2b8955a525b1d3211ef3385421cdb69","mode":"go"}},{"id":1,"method":"workspace/symbol","params":{"query":"dir:/src/github.com/kubernetes/kubernetes/federation/pkg/dnsprovider/ startChangeset","limit":100}},{"id":2,"method":"shutdown"},{"method":"exit"}]' http://localhost:3080/.api/xlang/workspace/symbol|python -mjson.tool
[
    {
        "id": 0,
        "jsonrpc": "2.0",
        "result": {
            "definitionProvider": true,
            "hoverProvider": true,
            "referencesProvider": true
        }
    },
    {
        "id": 0,
        "jsonrpc": "2.0",
        "result": []
    },
    {
        "id": 0,
        "jsonrpc": "2.0",
        "result": {}
    },
    null
]
```

I am seeing it turn up empty.

Also, I am not sure the client that constructs the workspace/symbol or textDocument/symbol query will necessarily know how to construct the directory path for `dir:xyz` in the query. It knows the import path. The mapping from import path to directory path is probably usually straightforward, but not always (you need to know if it's a Go stdlib package in GOROOT or a normal package in GOPATH, and in the case of multiple GOPATHs, you need to know which GOPATH it's in). It also seems as though in the end the matching is based on package import path, not directory. So, can the client just specify `pkg:github.com/foo/bar/baz MySymbol` instead of `dir:/src/github.com/foo/bar/baz MySymbol`?

slimsag:
@sqs 

> Would you expect this to work on Sourcegraph?

No, the directory path is relative to the workspace root path. Your query would be `dir:federation/pkg/dnsprovider/ startChangeset`

sqs:
@slimsag I'm not seeing that return any results either.

```
$ curl -sS --data '[{"id":0,"method":"initialize","params":{"rootPath":"git://github.com/kubernetes/kubernetes?6184ebdcb2b8955a525b1d3211ef3385421cdb69","mode":"go"}},{"id":1,"method":"workspace/symbol","params":{"query":"dir:federation/pkg/dnsprovider/ startChangeset","limit":100}},{"id":2,"method":"shutdown"},{"method":"exit"}]' http://localhost:3080/.api/xlang/workspace/symbol|python -mjson.tool
[
    {
        "id": 0,
        "jsonrpc": "2.0",
        "result": {
            "definitionProvider": true,
            "hoverProvider": true,
            "referencesProvider": true
        }
    },
    {
        "id": 0,
        "jsonrpc": "2.0",
        "result": []
    },
    {
        "id": 0,
        "jsonrpc": "2.0",
        "result": {}
    },
    null
]
```

slimsag:
@sqs try `dir:federation/pkg/dnsprovider testDNSZone` Presumably there is no symbol called "startChangeset" inside of the dnsprovider package? (lol)

```
$ curl -sS --data '[{"id":0,"method":"initialize","params":{"rootPath":"git://github.com/kubernetes/kubernetes?6184ebdcb2b8955a525b1d3211ef3385421cdb69","mode":"go"}},{"id":1,"method":"workspace/symbol","params":{"query":"dir:federation/pkg/dnsprovider testDNSZone","limit":100}},{"id":2,"method":"shutdown"},{"method":"exit"}]' http://localhost:3080/.api/xlang/workspace/symbol | jq
[
  {
    "id": 0,
    "result": {
      "hoverProvider": true,
      "definitionProvider": true,
      "referencesProvider": true
    },
    "jsonrpc": "2.0"
  },
  {
    "id": 0,
    "result": [
      {
        "containerName": "dnsprovider",
        "kind": 14,
        "location": {
          "range": {
            "end": {
              "character": 10,
              "line": 50
            },
            "start": {
              "character": 0,
              "line": 50
            }
          },
          "uri": "git://github.com/kubernetes/kubernetes?6184ebdcb2b8955a525b1d3211ef3385421cdb69#federation/pkg/dnsprovider/dns_test.go"
        },
        "name": "testDNSZone"
      }
    ],
    "jsonrpc": "2.0"
  },
  {
    "id": 0,
    "result": {},
    "jsonrpc": "2.0"
  },
  null
]
```

sqs:
You are 100% right. My bad! :smile: 

Times on my machine for the BEFORE query `testDNSZone` and the AFTER query `dir:federation/pkg/dnsprovider testDNSZone` (using the new filter)

First run:
8130msec -> 1900msec

After it has run 10 times:
BEFORE 340msec -> AFTER 25msec

LGTM :party: :balloon: :racehorse: :speedboat: :ship: :ship: :party: :balloon: :racehorse: :speedboat: :ship: :ship:

slimsag:
> Also, I am not sure the client that constructs the workspace/symbol or textDocument/symbol query will necessarily know how to construct the directory path for dir:xyz in the query. It knows the import path. The mapping from import path to directory path is probably usually straightforward, but not always (you need to know if it's a Go stdlib package in GOROOT or a normal package in GOPATH, and in the case of multiple GOPATHs, you need to know which GOPATH it's in). It also seems as though in the end the matching is based on package import path, not directory. So, can the client just specify pkg:github.com/foo/bar/baz MySymbol instead of dir:/src/github.com/foo/bar/baz MySymbol?

For others onlooking, we talked about this a bit on Slack. I'm still going to look into it more tomorrow. There are some interesting problems to tackle here (and not just for Go).

> It also seems as though in the end the matching is based on package import path, not directory.

AFAIK, this is not true / the end matching is based on directory NOT import path. See also the tests which do test some subdirectories. The reason this is true is because we join `h.Init.RootImportPath` into the directory matcher ([see here](https://github.com/sourcegraph/go-langserver/pull/15/files#diff-41a3fbc5d9dbf547085a3235ea31fd2aR233)) with the direct intent of using that newly constructed path to make path equality tests _based on the import path_ ([see here](https://github.com/sourcegraph/go-langserver/pull/15/files#diff-41a3fbc5d9dbf547085a3235ea31fd2aR271)). That is, the fact that these coincidently align with import paths is an implementation detail and not a part of the query specification.

If any of this is wrong, I'll follow up with more PRs. Let's consider this correct until proven wrong, and work towards a better immediate future! :)

I'll squash + merge

PR #17, opened 2016-11-02
langserver: limit workspace/reference to 1/4 CPU by default + add WORKSPACE_REFERENCE_PARALLELISM env var

slimsag:
This limits workspace/reference to using only one fourth of the CPU by default. This is because, currently, it is quite an intensive operation. Typically the operation is only ran in the background, so it running slower is OK but not when it is a detriment to every other LSP operation (which is often user-invoked).

Running with `WORKSPACE_REFERENCE_PARALLELISM=8` will retain the old behavior prior to this change.

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -6,7 +6,10 @@ import (
 	"go/build"
 	"go/token"
 	"log"
+	"os"
+	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


langserver/workspace_refs.go
```diff
@@ -33,8 +36,23 @@ func (h *LangHandler) handleWorkspaceReference(ctx context.Context, conn JSONRPC
 		pkgPat = h.init.RootImportPath + "/..."
 	}
 
+	var paralellism int
```
beyang: s/paralellism/parallelism/ :P

langserver/workspace_refs.go
```diff
@@ -33,8 +36,23 @@ func (h *LangHandler) handleWorkspaceReference(ctx context.Context, conn JSONRPC
...
+	e := os.Getenv("WORKSPACE_REFERENCE_PARALLELISM")
```
beyang: Thanks for fixing this @slimsag :)

I worry about the use of ad hoc environment variables like this, as they affect behavior implicitly and they're easy to overlook. Over time, more might be added and then it becomes difficult to even list all the env variables that can affect the app. Could we move this env variable to a separate package or file (e.g., `config.go`) so they are obviously configuration points and in one place?

langserver/workspace_refs.go
slimsag: _Shoooot._ That word is so hard to spell. Fixed in f8788d94052a4cbc8b5bd6cf590c451bb6e1d80b

langserver/workspace_refs.go
slimsag: Yeah.. I actually think it is a bad idea to use environment variables to tweak things like this anyway. If you think about it, the person modifying the environment has to choose between slow and fast with no good in-between. In an ideal world, work is scheduled based on what work the language server itself is doing and what global resources are allocated to it as a whole. This is harder to implement, so.. e95b4c79df487d9aa01c4bcfe5a4f6047840c423 as you suggest for now.

PR #18, opened 2016-11-03
Fix def range

nicot:
Update def range to be a position instead of a range. This fixes a bug
where a hover tooltip would only show one line when pressing cmd/ctrl,
instead of the entire definition.

This is the behavior of the most popular Go language server.


uforic:
If this is the behavior of the most popular Go language server, LGTM.

PR #19, opened 2016-11-04
lsp: add doc symbol capability

uforic:
@slimsag did the hard work, I get to steal all the glory by adding the capability.


slimsag:
LGTM.

But what does this do? I mean, if I hover over `Router` here it displays docs already: https://sourcegraph.com/github.com/gorilla/mux/-/blob/mux.go#L39

maybe it's for VS Code which I haven't tried?

keegancsmith:
This change advertises the fact that we support `textDocument/documentSymbol`, which we do! Sourcegraph ignores this list at the moment, but vscode does not.

uforic:
Yep! For VSCode... or in this case, Sublime Text 3: 
https://cl.ly/123a2m1X3L3p

PR #21, opened 2016-11-04
Add panic handler to worker goroutines

keegancsmith:
Currently we are getting panics sometimes running workspace/symbols due to a bug
in `go/doc`. https://github.com/golang/go/issues/17788

This change prevents panics in our pkg collector goroutines taking the whole
server down. In the future we may want to have a better approach so we don't
accidently ignore panics, but for now this lets us move forward.

Fixes https://github.com/sourcegraph/sourcegraph/issues/2053


slimsag:
Sorry for letting this change get stale, I think rebasing should be straightforward.

PR #24, opened 2016-11-07
langserver: significantly reduce the amount of typechecking for workspace/reference

slimsag:
Prior to this change `workspace/reference` would individually typecheck each package
in the workspace that it intended to list external references for. This meant that
e.g. if a workspace had two packages `pkg/a` and `pkg/b` which both use `net/http`
that `net/http` and all of it's dependencies would have to be typechecked twice (once
for each input package, respectively).

After this change, we no longer cache typechecking results for workspace/reference
(in fact, it is not entirely useful to cache them due to workspace/reference being
a bad indicator of real direct user-initiated queries) and instead typecheck all
packages we intend to list external references of, together. This means that in the
prior example, `net/http` and it's dependencies will only be typechecked once. This
is possible because `go/loader.Config.Import` is smart enough to reuse existing
dependency typecheckign results.

The end effect is that a large repository like e.g. kubernetes has a quite large
perf gain in terms of `workspace/reference` typechecking, going from many minutes
(15+ before I killed it) to just 9-10s.

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"go/build"
+	"go/parser"
 	"go/token"
+	"go/types"
 	"log"
 	"runtime"
 	"sort"
@@ -13,6 +15,7 @@ import (
 	"sync"
 
 	"golang.org/x/tools/go/buildutil"
+	"golang.org/x/tools/go/loader"
 
 	"github.com/neelance/parallel"
 	opentracing "github.com/opentracing/opentracing-go"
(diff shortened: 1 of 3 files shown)


langserver/workspace_refs.go
```diff
@@ -49,14 +52,38 @@ func (h *LangHandler) handleWorkspaceReference(ctx context.Context, conn JSONRPC
...
+		fset = token.NewFileSet()
+		pkgs []string
+	)
+	for pkg := range buildutil.ExpandPatterns(bctx, []string{pkgPat}) {
+		// Ignore any vendor package so we can avoid scanning it for external
+		// references, per the workspace/reference spec. This saves us a
+		// considerable amount of work.
+		bpkg, err := bctx.Import(pkg, rootPath, build.FindOnly)
+		if err != nil && !isMultiplePackageError(err) {
+			log.Printf("skipping possible package %s: %s\n", pkg, err)
```
keegancsmith: don't need the `\n` for `log.Printf`

langserver/workspace_refs.go
```diff
@@ -77,65 +104,90 @@ func (h *LangHandler) handleWorkspaceReference(ctx context.Context, conn JSONRPC
...
-	buildPkg, err := bctx.Import(pkg, rootPath, 0)
-	if err != nil {
-		if !(strings.Contains(err.Error(), "no buildable Go source files") || strings.Contains(err.Error(), "found packages") || strings.HasPrefix(pkg, "github.com/golang/go/test/")) {
-			return fmt.Errorf("skipping possible package %s: %s\n", pkg, err)
-		}
-		return nil
+	// Configure the loader.
+	var typeErrs []error
+	conf := loader.Config{
+		Fset: fset,
```
keegancsmith: so is reusing the `fset` the magic that prevents duplicate typechecking?

langserver/workspace_refs.go
```diff
@@ -77,65 +104,90 @@ func (h *LangHandler) handleWorkspaceReference(ctx context.Context, conn JSONRPC
...
-		// which saves us a a lot of work.
-		return nil
+	for _, path := range pkgs {
+		conf.Import(path)
 	}
 
-	pkgInWorkspace := func(path string) bool {
-		return PathHasPrefix(path, h.init.RootImportPath)
+	// Load and typecheck the packages.
+	prog, err = conf.Load()
```
slimsag: @keegancsmith the magic is actually here. This only gets called once, for all packages (see `conf.Import(path)` above) rather than once for each package as it was before.

Basically, keeping only one `loader.Config` for all your request-scoped work is a good idea :)

keegancsmith: Oh I totally misread this :) I thought `externalRefsTypeCheck` was called in a loop, but I see now it isn't. Nice!

PR #28, opened 2016-11-09
langserver: properly make workspace/symbol `dir:` filter operate on directory

slimsag:
In majority of cases, package path and directory path align (relative to a Go
package directory), however this is not always true. For example, `github.com/golang/go`
has import paths `net/http` but you would expect `dir:src/net/http` to function. I believe there are also other cases of this, outside of just this repo.


langserver/symbol.go
```diff
@@ -417,3 +422,9 @@ func parseDir(fset *token.FileSet, bctx *build.Context, path string, filter func
 
 	return
 }
+
+func maybeLogImportError(pkg string, err error) {
+	if !(strings.Contains(err.Error(), "no buildable Go source files") || strings.Contains(err.Error(), "found packages") || strings.HasPrefix(pkg, "github.com/golang/go/test/")) {
```
sqs: check for err types build.NoGoError, build.MultiplePackageError for type safety

sqs:
LGTM

How come the tests didn't catch this? Is there a test case or a new type of test we should have that would catch this?

slimsag:
Closing in favor of https://github.com/sourcegraph/go-langserver/pull/30 -- I will follow up on your question there. Not yet ready.

PR #30, opened 2016-11-09
langserver: properly make workspace/symbol `dir:` filter operate on directory

slimsag:
https://github.com/sourcegraph/go-langserver/pull/28 incorrectly tried to apply a fix that would make `workspace/symbol` with `dir:src/net/http` for github.com/golang/go work. The reason it was incorrect, was that it was special casing handling of that repository (effectively) inside the language server, here. This became apparent as I tried to add a proper test for this case.

The more correct fix is to handle this in our build server layer. This PR includes minor code cleanup, and exposing our workspace/symbol query parsing mechanisms in order to support fixing this in our build server by rewriting the `dir:` filter portion of queries.

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -2,6 +2,7 @@ package langserver
 
 import (
 	"context"
+	"fmt"
 	"go/ast"
 	"go/build"
 	"go/doc"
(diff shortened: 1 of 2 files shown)


langserver/symbol.go
```diff
@@ -32,6 +33,38 @@ type Query struct {
 	Tokens    []string
 }
 
+// String converts the query back into a logically equivilent, but not strictly
```
keegancsmith: spelling nit: equivalent

langserver/symbol.go
```diff
@@ -32,6 +33,38 @@ type Query struct {
...
+	for _, token := range q.Tokens {
+		s = queryJoin(s, token)
+	}
+	return s
+}
+
+// queryJoin joins the strings into "<s><space><e>" ensuring there is no
+// trailing or leading whitespace at the end of the string.
+func queryJoin(s, e string) string {
+	return strings.TrimSpace(strings.Join([]string{s, e}, " "))
```
keegancsmith: nit: why don't you just use `s + " " + e` instead of `strings.Join`?

langserver/symbol.go
slimsag: I've enabled [Sublime's spell checker](https://forum.sublimetext.com/t/auto-spell-check-sublime-3/12930) so I'll stop making these stupid mistakes :)

Issue #11, opened 2016-10-27
When run locally, document/references returns references that aren't in the initial root_path


uforic:
When debugging the go language server locally, I noticed that a references request pulled back refs that weren't in the root_path of the project.

Step 1: Initialize language server, notice the rootPath
`
language server initialized: [{"jsonrpc": "2.0", "id": 1, "result": {"capabilities": {"workspaceSymbolProvider": true, "textDocumentSync": 1, "definitionProvider": true, "hoverProvider": true, "referencesProvider": true}}}, {"request": {"params": {"capabilities": {}, "processId": 5506, "rootPath": "file:///Users/mattfs/gopath/src/sourcegraph.com/sourcegraph/sourcegraph"}, "jsonrpc": "2.0", "id": 1, "method": "initialize"}}]
`

Step 2: Make a references call, was on net.Listener in server.go:
`
{'params': {'includeDeclaration': True, 'position': {'line': 43, 'character': 32}, 'textDocument': {'uri': 'file:///Users/mattfs/gopath/src/sourcegraph.com/sourcegraph/sourcegraph/pkg/gitserver/server.go'}}, 'jsonrpc': '2.0', 'id': 3, 'method': 'textDocument/references'}
`

Step 3: A sampling of results returned (over 130 results):
`
file:///Users/mattfs/gopath/src/github.com/chzyer/readline/remote.go
`
`
file:///Users/mattfs/gopath/src/github.com/rogpeppe/rog-go/loopback/net.go
`

Notice how none of these are vendored (how I know they aren't coming from Sourcegraph), and they all happen to be in my GOPATH.

Is it possible we are pulling all references to a given def that are in the GOPATH, not just the ones in the doc_root of the project we are in?

keegancsmith:
This is actually the same behaviour as any other go plugin. (go guru searches the whole GOPATH). However, I agree that this is surprising behaviour for find _local_ references.

I am guessing we can do one (or more) of three things:
1. Limit results to `rootPath`.
2. Order results intelligently.
3. Wait until we references searches global as well (will still need 2).

(1) should be pretty easy to implement. (2) for just boosting results in rootPath should also be very easy.

cc @sqs

sqs:
Iirc this is filed somewhere already or mentioned in code TODO. Agreed we should do (1) below now.

uforic:
[...]

PR #31, opened 2016-11-10
references: Cache the reverse import graph

keegancsmith:
We compute this every time we answer a references query and for large workspaces
it dominates the time. However, it is static W.R.T. the workspace. In the future
when we start doing cache invalidation, this will need to be updated.

On my machine it takes subsequent references queries from 20s to 3s.

Part of #11

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -31,7 +31,11 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	}
 
 	bctx := h.OverlayBuildContext(ctx, h.defaultBuildContext(), !h.init.NoOSFileSystemAccess)
-	_, rev, _ := importgraph.Build(bctx)
+	h.importGraphOnce.Do(func() {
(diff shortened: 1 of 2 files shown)


langserver/references.go
```diff
@@ -31,7 +31,10 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	}
 
 	bctx := h.OverlayBuildContext(ctx, h.defaultBuildContext(), !h.init.NoOSFileSystemAccess)
-	_, rev, _ := importgraph.Build(bctx)
+	h.importGraphOnce.Do(func() {
+		_, rev, _ := importgraph.Build(bctx)
+		h.importGraph = rev
```
keegancsmith: We ignored the errors previously. I assume we ignore the error since we are doing best effort, so we just ignore code which doesn't compile. I'll add in a comment.

keegancsmith:
> I assume this change is already covered by existing tests.

yup

PR #32, opened 2016-11-10
refs: Order results by shared directories of query

keegancsmith:
Now references in the same package will be listed first, followed by
subpackages, then other packages. For other packages, if they share more
directories with the input than another package, it will be listed before. See
the test case for a concrete example.

Part of #11

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -8,6 +8,7 @@ import (
 	"go/parser"
 	"go/token"
 	"go/types"
+	"path"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


langserver/references.go
```diff
@@ -240,21 +241,54 @@ func sameObj(x, y types.Object) bool {
...
+	D []int
+}
 
 func (l locationList) Less(a, b int) bool {
-	if l[a].URI != l[b].URI {
-		return l[a].URI < l[b].URI
+	if l.D[a] != l.D[b] {
+		return l.D[a] > l.D[b]
+	}
+	if x, y := path.Dir(l.L[a].URI), path.Dir(l.L[b].URI); x != y {
```
keegancsmith: we could precompute this to make this faster, but the simplicity of this approach makes it more readable. I may in fact remove the precomputation of `l.D` to simplify that code as well if someone things I should.

PR #37, opened 2016-11-23
langserver: Run empty workspace/symbol on initialize

keegancsmith:
This should get a bunch of files typechecked on startup, and hopefully hide some
latency from users.


beyang:
LGTM, but looks like this trips the go race detector somehow

keegancsmith:
Ok, will fix race conditions, even if benign.

PR #46, opened 2016-12-03
workspace/reference: improve testing + fix serious bug with multiple files

slimsag:
- Improve testing of `workspace/reference` generally.
- Do not emit `End` ranges that are bizarrely off or broken. Instead, emit `Start == End` and leave it as a future TODO (we do not need them right now).
- Fix an off-by-one error in the resulting `start`/`end` characters of the ranges.
- Fix a serious bug that causes the emitted filename/position for references to be VeryWrong™ / random (and add a test case).

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -234,7 +234,8 @@ func (h *LangHandler) externalRefsFromPkg(ctx context.Context, bctx *build.Conte
 			Name:          defName,
 			ContainerName: defContainerName,
 			URI:           "file://" + defPkg.Dir,
-			Location:      goRangeToLSPLocation(fs, token.Pos(r.Position.Offset), token.Pos(r.Position.Offset+len(defName)-1)),
+			// TODO: This code is EXTREMELY invalid!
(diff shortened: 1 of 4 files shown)


langserver/workspace_refs.go
```diff
@@ -234,8 +234,7 @@ func (h *LangHandler) externalRefsFromPkg(ctx context.Context, bctx *build.Conte
 			Name:          defName,
 			ContainerName: defContainerName,
 			URI:           "file://" + defPkg.Dir,
-			// TODO: This code is EXTREMELY invalid!
-			Location: goRangeToLSPLocation(fs, token.Pos(r.Position.Offset)+1, token.Pos(r.Position.Offset+len(defName)-1)),
+			Location:      goRangeToLSPLocation(fs, r.Pos, r.Pos), // TODO: internal/refs doesn't generate end positions
```
keegancsmith: > We previously would convert a `token.Position`s `Offset` field arbitrarily to `token.Pos`.
This was probably because I at one point in time thought this was correct (because it works
on singular files, like in all of our test cases), but it is entirely NOT correct. It's like
arbitrarily casting from one unit to another, e.g. MPH to KPH, and disregarding any conversion
logic that is needed.

Love how simple the fix is. Nice! And thanks for all the build up to the fix in the preceding commits.

PR #47, opened 2016-12-05
RFC: Cache extension

keegancsmith:
This is intended to kick of the discussion of what the cache extension will look like, which we want to use for future golang/php improvements.

This is a very simplistic cache interface, which doesn't take into account a bunch of properties a more complex implementation would have. It also does not specify any constraints on value/key sizes, which we may want to indicate somehow since the backing cache store will have constraints.

### DIFF
--- a/pkg/lspext/cacheext.go
+++ b/pkg/lspext/cacheext.go
@@ -2,14 +2,14 @@ package lspext
 
 import "encoding/json"
 
-// CacheGetParams is the input for 'cache/get'. The response is a
+// CacheGetParams is the input for 'xcache/get'. The response is a
 // 'CacheItem'. This cache is global to a language server. IE it is shared
 // amongst workspaces, but not amongst different language server types.
 type CacheGetParams struct {
 	Key string `json:"key"`
 }
 
-// CacheItem is the response for 'cache/get'.
+// CacheItem is the response for 'xcache/get'.
 type CacheItem struct {
 	// Value is the value stored in the cache. It is a *json.RawMessage
 	// since the type in the spec is `any`. It is nil if the value is not
@@ -17,7 +17,7 @@ type CacheItem struct {
 	Value *json.RawMessage `json:"value,omitempty"`
 }
 
-// CacheSetParams is the input for 'cache/set'. It is a notify method, so
+// CacheSetParams is the input for 'xcache/set'. It is a notify method, so
 // does not have a response.
 type CacheSetParams struct {
 	Key string `json:"key"`
--- a/pkg/lspext/cacheext.go
+++ b/pkg/lspext/cacheext.go
@@ -1,23 +1,27 @@
 package lspext
 
-// CacheGetParams is the input for 'cache/xget'. The response is a
+import "encoding/json"
+
+// CacheGetParams is the input for 'cache/get'. The response is a
 // 'CacheItem'. This cache is global to a language server. IE it is shared
 // amongst workspaces, but not amongst different language server types.
 type CacheGetParams struct {
 	Key string `json:"key"`
 }
 
-// CacheItem is the response for 'cache/xget'.
+// CacheItem is the response for 'cache/get'.
 type CacheItem struct {
-	// Ok is true if the item was successfuly retrieved from the cache.
-	Ok bool `json:"ok"`
-	// Value is the value stored in the cache. It is present if Ok is true.
-	Value string `json:"value,omitempty"`
+	// Value is the value stored in the cache. It is a *json.RawMessage
+	// since the type in the spec is `any`. It is nil if the value is not
+	// in the cache.
+	Value *json.RawMessage `json:"value,omitempty"`
 }
 
-// CacheSetParams is the input for 'cache/xset'. It is a notify method, so
+// CacheSetParams is the input for 'cache/set'. It is a notify method, so
 // does not have a response.
 type CacheSetParams struct {
-	Key   string `json:"key"`
-	Value string `json:"value"`
+	Key string `json:"key"`
+	// Value is the same as CacheItem.Value. Note when setting Value
+	// cannot be nil.
+	Value *json.RawMessage `json:"value"`
 }
--- a/pkg/lspext/cacheext.go
+++ b/pkg/lspext/cacheext.go
@@ -1,0 +1,23 @@
+package lspext
+
+// CacheGetParams is the input for 'cache/xget'. The response is a
+// 'CacheItem'. This cache is global to a language server. IE it is shared
+// amongst workspaces, but not amongst different language server types.
+type CacheGetParams struct {
+	Key string `json:"key"`
+}
+
+// CacheItem is the response for 'cache/xget'.
+type CacheItem struct {
+	// Ok is true if the item was successfuly retrieved from the cache.
+	Ok bool `json:"ok"`
+	// Value is the value stored in the cache. It is present if Ok is true.
+	Value string `json:"value,omitempty"`
+}
+
+// CacheSetParams is the input for 'cache/xset'. It is a notify method, so
+// does not have a response.
+type CacheSetParams struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
--- a/pkg/lsp/service.go
+++ b/pkg/lsp/service.go
@@ -21,8 +21,8 @@ type ClientCapabilities struct {
 	// textDocument/xcontent. This is a Sourcegraph extension.
 	XContentProvider bool `json:"xcontentProvider,omitempty"`
 
-	// XCacheProvider indicates the client provides support for cache/xget
-	// and cache/xset. This is a Sourcegraph extension.
+	// XCacheProvider indicates the client provides support for cache/get
+	// and cache/set. This is a Sourcegraph extension.
 	XCacheProvider bool `json:"xcacheProvider,omitempty"`
 }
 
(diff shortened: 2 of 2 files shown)


pkg/lspext/cacheext.go
```diff
@@ -0,0 +1,23 @@
...
+	Ok bool `json:"ok"`
+	// Value is the value stored in the cache. It is present if Ok is true.
+	Value string `json:"value,omitempty"`
+}
+
+// CacheSetParams is the input for 'cache/xset'. It is a notify method, so
+// does not have a response.
+type CacheSetParams struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
```
felixfbecker: I don't know how this would look like in Go, but `value` should be any type, not just a string. The JS/TS LS will likely want to store objects directly.

pkg/lspext/cacheext.go
```diff
@@ -0,0 +1,23 @@
+package lspext
+
+// CacheGetParams is the input for 'cache/xget'. The response is a
```
keegancsmith: I don't mind, sure latter sounds good to me.

pkg/lspext/cacheext.go
```diff
@@ -0,0 +1,23 @@
...
+// 'CacheItem'. This cache is global to a language server. IE it is shared
+// amongst workspaces, but not amongst different language server types.
+type CacheGetParams struct {
+	Key string `json:"key"`
+}
+
+// CacheItem is the response for 'cache/xget'.
+type CacheItem struct {
+	// Ok is true if the item was successfuly retrieved from the cache.
+	Ok bool `json:"ok"`
```
keegancsmith: I think using errors for cache misses is mixing up exceptional circumstances with something you should expect to happen (ie a cache miss is not a rare event). We may be coming at this with different expectations, but I think of a cache as something which we expect to be missing values.

felixfbecker: Definitely. But wouldn't omitting `value` / returning `null` for this be enough? That's what the JS Map does when the entry does not exist, it returns `undefined`

pkg/lspext/cacheext.go
keegancsmith: Sure we can use any. In Go it would be an `interface{}` or for something to avoid serialization a `json.RawMessage`. The reason I picked a string is it doesn't hide the fact that your cache will have to serialize anyways, which better communicates to the developer the tradeoffs. Not sure what your thoughts are on that.

keegancsmith:
> A client may at any time clear cache items to save space. The server must at all times be able to fallback to the real source of data, even directly after a cache/set request.

I'd prefer to rely on a cache just being something which evicts things it thinks are not used anymore.  How do we fallback to real source data in the cache server, it isn't aware how data is generated. The LSP should instead do a Get, and if that says it doesn't have the key, then do the computation.

> The namespace of the cache key is shared across projects, but not across language servers.

Agreed, is my docstring not clear enough for that?

> As a reference, I would like the cache namespace to mirror the API of JavaScript's Map. I think it would be beneficial and very low effort to also add cache/clear, cache/delete, cache/has and cache/entries (for both real usage and debugging purposes).

As I somewhat mentioned previously, I agree it is low effort to add those features to the API, but it leads to language server developers relying on features it doesn't really need but will lead to bugs + poorly performing code. Examples:
* Listing keys is a slow operation in a distributed cache.
* You can have more than one PHP replica running, which means one can clear the cache and trample over the expected state of another PHP replica that is running. I can give more examples if ya like.

> It needs to be specified what happens when a cache key is requested that does not exist / has been deleted. Should it return an error with a new error code or should it return undefined value, like JS Map?

That is the `Ok` field I added. I can make this clearer by expanding on doc strings, or write up a spec. For now I think we can just debate it on this PR :)

> What replacement algorithm should the client use if the cache is full? FIFO? LRU? Should we spec this, or can we just leave this to the client implementation?

I think we should leave it to the client implementation. The only expectation we should set is that keys can be dropped, like all caches.

felixfbecker:
> I'd prefer to rely on a cache just being something which evicts things it thinks are not used anymore. How do we fallback to real source data in the cache server, it isn't aware how data is generated. The LSP should instead do a Get, and if that says it doesn't have the key, then do the computation.

That's what I meant (with server I meant the LS).

> Agreed, is my docstring not clear enough for that?

Overlooked it - the first two bullet points were just some thoughts I had :)

> As I somewhat mentioned previously, I agree it is low effort to add those features to the API, but it leads to language server developers relying on features it doesn't really need but will lead to bugs + poorly performing code. Examples:
Listing keys is a slow operation in a distributed cache.
You can have more than one PHP replica running, which means one can clear the cache and trample over the expected state of another PHP replica that is running. I can give more examples if ya like.

Okay, good points about clear and entries. What about has and delete? I imagine LS will want to do some sort of manual cache invalidation at some point

> That is the Ok field I added. 

Okay, figured that later. I think just returning a null/undefined value would be simpler though, wdyt?

> I think we should leave it to the client implementation. The only expectation we should set is that keys can be dropped, like all caches.

I agree.

pkg/lspext/cacheext.go
keegancsmith: Sure we can do that as well. I did it this way for three reasons:
* Explicit.
* Avoid nulls.
* Avoid empty values (eg `""`) having a special meaning.

For example in some languages the JSON parser makes it difficult to distinguish between a value that isn't present vs empty (languages that generally attempt to deal with null/undefined much better than dynamic languages do)

keegancsmith:
> Okay, good points about clear and entries. What about has and delete? I imagine LS will want to do some sort of manual cache invalidation at some point

What is a usecase of just checking if a key is set? + There are no guarantees between doing a has and a get that the key is still present.
What is the use case of relying on deleting a value in a global cache? Especially when you think about multiple replicas running / analysing different commits at the same time. Allowing a delete seems like it would encourage people to use the cache in a way that it isn't a simple cache anymore.

pkg/lspext/cacheext.go
felixfbecker: I see your point, but it would require the JS/TS LS for example to `JSON.stringify()` the object on the server side, basically double-encoding it. This also means that you will have to double-escape everything, so it will result in more bytes being sent over the wire.

For my own use case, it makes no difference. I will use PHP's own serialization to keep class information.

felixfbecker:
Good points. I just found a use case for `clear` though - between versions of an LS the index format will likely change, so the cache format has to change too (the items still in the cash have a different format now, or are missing new properties). This requires a cache clear & rebuild. But it would probably be best to let the client nuke the cache on every update (like it will for every uninstall), instead of forcing the LS to implement migration logic. Probably the best idea to keep it at `get`/`set`.

What do you think of using `ok` vs returning `null`/`undefined`? It would remove the transitive dependency between `value` and `ok`

pkg/lspext/cacheext.go
keegancsmith: Aah, I now lean towards doing `any`. I can see how a string always requires escaping, so that means we would do "encoding" twice. I'll update to `any`.

keegancsmith:
> Good points. I just found a use case for clear though - between versions of an LS the index format will likely change, so the cache format has to change too (the items still in the cash have a different format now, or are missing new properties). This requires a cache clear & rebuild. But it would probably be best to let the client nuke the cache on every update (like it will for every uninstall), instead of forcing the LS to implement migration logic. Probably the best idea to keep it at get/set.

In a distributed environment, you can never guarantee that everything is running at the exact same version all the time, so I'd rather not introduce those sort of requirements. Instead if an LS wants to change the format of how something is serialized, they should include some sort of versioning in the cache key (like one normally does for that sort of concern).

> What do you think of using ok vs returning null/undefined? It would remove the transitive dependency between value and ok

One point I'd argue is if `value` is an `any`, then should we allow someone to do `set("foo", null)` since `null` is an `any`? The `ok` makes it explicitly clear that we retrieved something. But, I have seen plenty of cache implementations that rely on nullness to communicate it, we should just explicitly say you can't cache the null value.

felixfbecker:
> In a distributed environment, you can never guarantee that everything is running at the exact same version all the time, so I'd rather not introduce those sort of requirements. Instead if an LS wants to change the format of how something is serialized, they should include some sort of versioning in the cache key (like one normally does for that sort of concern).

So an example cache key in my case would look like this:
```
1~symfony/symfony:2.4.1
```
where `1` is the incremental cache format version
when you update the LS, what happens to the old cache items - will they just get deleted after a client-specified time?

> One point I'd argue is if value is an any, then should we allow someone to do set("foo", null) since null is an any? The ok makes it explicitly clear that we retrieved something. But, I have seen plenty of cache implementations that rely on nullness to communicate it, we should just explicitly say you can't cache the null value.

True, so let's just document `value` can be any type except `null`.

keegancsmith:
> when you update the LS, what happens to the old cache items - will they just get deleted after a client-specified time?

they can disappear at anytime, so it is upto the client. But if it is backed by something like memcached they will just disappear once they can no longer fit in memory and they get evicted. Not wasteful at all since either memcached has too much memory and keeps things that aren't used around for too long or they will be evicted not too long after they key isn't used anymore.


On another note we should have some concrete uses for this stuff before proposing upstream, just to prove there worth / give us flexibility to adjust them. But great to discuss all this now :)

felixfbecker:
I meant we should still use a prefix, but like `xcache/get`, not `cache/xget`. I think the prefix is still important if we don't want to break our own LS with the standard proposal.

keegancsmith:
I am not gonna merge this in since it was just an RFC. When we actually want this behaviour we can use this as a reference / revive the branch.

PR #48, opened 2016-12-06
all: implement workspace/reference v2

slimsag:
According to the new specification at https://github.com/sourcegraph/language-server-protocol/pull/8

Please do not merge until that PR is.

### DIFF
--- a/langserver/definition.go
+++ b/langserver/definition.go
@@ -69,12 +69,8 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn JSONRPC2Conn, 
 		l.Location.Range.End = l.Location.Range.Start
 
 		// Determine metadata information for the node.
-		//rootFile := pkg.Files[0] // TODO: is this correct?
-		def, err := refs.DefInfo(pkg.Pkg, &pkg.Info, pathEnclosingInterval, node.Pos())
-		if err != nil {
-			// TODO: tracing
-			log.Println("refs.DefInfo:", err)
-		} else {
+
+		if def, err := refs.DefInfo(pkg.Pkg, &pkg.Info, pathEnclosingInterval, node.Pos()); err == nil {
 			symDesc, err := defSymbolDescriptor(bctx, rootPath, *def)
 			if err != nil {
 				// TODO: tracing
@@ -82,6 +78,9 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn JSONRPC2Conn, 
 			} else {
 				l.Symbol = *symDesc
 			}
+		} else {
+			// TODO: tracing
+			log.Println("refs.DefInfo:", err)
 		}
 		locs = append(locs, l)
 	}
--- a/langserver/definition.go
+++ b/langserver/definition.go
@@ -4,13 +4,31 @@ import (
 	"context"
 	"errors"
 	"go/ast"
+	"log"
 
+	"github.com/sourcegraph/go-langserver/langserver/internal/refs"
 	"github.com/sourcegraph/go-langserver/pkg/lsp"
+	"github.com/sourcegraph/go-langserver/pkg/lspext"
 	"github.com/sourcegraph/jsonrpc2"
 )
 
 func (h *LangHandler) handleDefinition(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) ([]lsp.Location, error) {
-	fset, node, _, pkg, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
+	res, err := h.handleXDefinition(ctx, conn, req, params)
+	if err != nil {
+		return nil, err
+	}
+	locs := make([]lsp.Location, 0, len(res))
+	for _, li := range res {
+		locs = append(locs, li.Location)
+	}
+	return locs, nil
+}
+
+func (h *LangHandler) handleXDefinition(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) ([]lspext.LocationInformation, error) {
+	rootPath := h.FilePath(h.init.RootPath)
+	bctx := h.OverlayBuildContext(ctx, h.defaultBuildContext(), !h.init.NoOSFileSystemAccess)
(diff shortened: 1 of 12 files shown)


langserver/workspace_refs.go
```diff
@@ -222,19 +209,19 @@ func (h *LangHandler) externalRefsFromPkg(ctx context.Context, bctx *build.Conte
...
+			Reference: goRangeToLSPLocation(fs, r.Pos, r.Pos), // TODO: internal/refs doesn't generate end positions
+			Symbol: map[string]interface{}{
+				// Standard fields.
+				"name":          defName,
+				"containerName": defContainerName,
+
+				// Non-standard fields.
+				"dir":    defPkg.Dir, // TODO: emit standard "file" instead?
+				"pkg":    defPkg.ImportPath,
+				"vendor": IsVendorDir(defPkg.Dir),
```
keegancsmith: shouldn't these fields live in meta?

langserver/workspace_refs.go
```diff
@@ -217,7 +217,6 @@ func (h *LangHandler) externalRefsFromPkg(ctx context.Context, bctx *build.Conte
 				ContainerName: defContainerName,
 				PackageName:   defPkg.ImportPath,
 				Meta: map[string]interface{}{
-					"dir":    defPkg.Dir, // TODO: emit standard "file" instead?
```
keegancsmith: surprised you didn't need to update tests with this commit?

Edit: I see you updated in a later commit

slimsag:
1. I've updated the `workspace/xreference` implementation to reflect the latest specification.
2. I've also added a `textDocument/xdefinition` implementation according to the spec.
3. I've also squashed this down into just a single commit for `workspace/xreferences` and `textDocument/xdefinition`.

The xreference impl has not changed much since @keegancsmith reviewed it, but the second commit / xdefinition implementation is new so I'll need to get a review on that. @keegancsmith I will ask someone else to review because I want to land this into master today, _but_ please leave comments etc here and I will follow up tomorrow morning.

langserver/definition.go
```diff
@@ -40,11 +58,32 @@ func (h *LangHandler) handleDefinition(ctx context.Context, conn JSONRPC2Conn, r
...
+		l := lspext.LocationInformation{
+			Location: goRangeToLSPLocation(fset, node.Pos(), node.End()),
+		}
 		// LSP expects a range to be of the entire body, not just of the
 		// identifier, so we pretend its just a position and not a range.
-		locs[i].Range.End = locs[i].Range.Start
+		l.Location.Range.End = l.Location.Range.Start
+
+		// Determine metadata information for the node.
+		//rootFile := pkg.Files[0] // TODO: is this correct?
```
beyang: remove commented-out code

langserver/definition.go
```diff
@@ -40,11 +58,32 @@ func (h *LangHandler) handleDefinition(ctx context.Context, conn JSONRPC2Conn, r
...
+		def, err := refs.DefInfo(pkg.Pkg, &pkg.Info, pathEnclosingInterval, node.Pos())
```
beyang: Since `def` is only used within the following if-else block, a clearer way to write this would be:

```
if def, err := refs.DefInfo(pkg.Pkg, &pkg.Info, pathEnclosingInterval, node.Pos()); err == nil {

} else 

}
```

langserver/workspace_refs.go
```diff
@@ -189,10 +180,6 @@ func (h *LangHandler) externalRefsFromPkg(ctx context.Context, bctx *build.Conte
 	}()
 	span.SetTag("pkg", pkg)
 
-	pkgInWorkspace := func(path string) bool {
```
beyang: If we now emit references to both internal and external definitions, you should change the name and docstring of this function.

pkg/lspext/lspext.go
```diff
@@ -1,47 +1,80 @@
...
-	// URI is the URI location of the symbol that is being referenced. If both
-	// Name and ContainerName are empty strings, it implies that the URI
-	// location is referenced but not a particular symbol (e.g. a Go import
-	// statement, a JS require() call, etc).
-	URI string `json:"uri"`
+	Reference lsp.Location     `json:"reference"`
+	Symbol    SymbolDescriptor `json:"symbol"`
+}
+
+type SymbolDescriptor struct {
```
beyang: Please add a docstring for this type and each of the fields.

pkg/lspext/lspext.go
```diff
@@ -1,47 +1,80 @@
...
-type WorkspaceReferenceParams struct {
-	Limit int `json:"limit"`
+type WorkspaceReferencesParams struct {
 }
 
+// ReferenceInformation is the array response type for the `workspace/xreferences` extension
+//
+// See: https://github.com/sourcegraph/language-server-protocol/blob/master/extension-workspace-reference.md
+//
 type ReferenceInformation struct {
```
beyang: Please add a docstring for each of these fields.

pkg/lspext/lspext.go
```diff
@@ -1,47 +1,80 @@
...
+	Symbol    SymbolDescriptor `json:"symbol"`
+}
+
+type SymbolDescriptor struct {
+	Name          string                 `json:"name,omitempty"`
+	Kind          lsp.SymbolKind         `json:"kind,omitempty"`
+	File          string                 `json:"file,omitempty"`
+	ContainerName string                 `json:"containerName,omitempty"`
+	Vendor        bool                   `json:"vendor,omitempty"`
+	Meta          map[string]interface{} `json:"meta,omitempty"`
```
beyang: If `SymbolDescriptor` itself is "metadata", having a `Meta` field inside it seems redundant. What about something like `Attributes`?

slimsag: That would be a spec change, too. But okay.

langserver/definition.go
slimsag: (hint: it was _not_ correct :) )

langserver/workspace_refs.go
slimsag: (done)

slimsag:
Also see: https://github.com/sourcegraph/sourcegraph/pull/2673

PR #50, opened 2016-12-07
langserver: Use global tracer if specified

keegancsmith:
Currently we send all telemetry events over jsonrpc. We can instead directly
send spans to our opentracing provider in production. This also helps with local
debugging/testing.

### DIFF
--- a/langserver/tracing.go
+++ b/langserver/tracing.go
@@ -23,6 +23,12 @@ func (h *HandlerCommon) InitTracer(conn *jsonrpc2.Conn) {
 		return
 	}
 
+	if t := opentracing.GlobalTracer(); t != nil {
+		h.tracer = t
+		h.tracerOK = true
+		return
+	}
+
 	t := tracer{conn: conn}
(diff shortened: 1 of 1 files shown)


langserver/tracing.go
```diff
@@ -23,6 +23,12 @@ func (h *HandlerCommon) InitTracer(conn *jsonrpc2.Conn) {
 		return
 	}
 
+	if t := opentracing.GlobalTracer(); t != nil {
+		h.tracer = t
```
neelance: Ugh, it seems very wrong that `HandlerCommon` has a `tracer` field. A tracer should always be a global singleton. Why is it built that way?

neelance: In which scenarios do we still need that way via `telemetry/event`?

keegancsmith: go-langserver running not on sourcegraph.com. Like vscode/sublime it will likely provide useful to have.

neelance: All right.

PR #55, opened 2016-12-13
langserver: Do not hold lock for FilePath

keegancsmith:
FilePath can currently panic, leading to deadlock since we never unlock the
mutex. We also do not need to hold the lock while using FilePath. In future, we
may make FilePath a function rather than a method since it doesn't access any
fields of h.

### DIFF
--- a/langserver/fs.go
+++ b/langserver/fs.go
@@ -91,8 +91,8 @@ func (h *HandlerShared) FilePath(uri string) string {
 func (h *HandlerShared) readFile(ctx context.Context, uri string) ([]byte, error) {
 	h.Mu.Lock()
 	fs := h.FS
-	path := h.FilePath(uri)
 	h.Mu.Unlock()
+	path := h.FilePath(uri)
 	contents, err := ctxvfs.ReadFile(ctx, fs, path)
 	if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


keegancsmith:
@sqs: I requested a review rather than just straight committing this simple change since I was interested why the FilePath function is part of handler. Is this since in the future we may adjust the path based on the init parameters?

sqs:
I am on mobile so I am going from memory, but it's to support the case where the workspace root is not st the root of the repo. If the root is file:///src/github.com/foo/bar/baz then the file path for file:///src/github.com/foo/bar/baz/qux.go is actuslly qux.go not baz/qux.go. Does that answer the question?

keegancsmith:
@sqs right now that function doesn't at all use the rootpath from init. It is a "pure" function.

sqs:
@keegancsmith Ah, now that I think of it, it previously used some of the handler fields to deal with that case (workspace root != repo root), but I simplified/changed how that works a while ago in such a way that no longer relies on FilePath. But I didn't follow through and actually make FilePath a func.

keegancsmith:
I'll update

keegancsmith:
It seems this function's name does not really make sense at the moment as well. For now I'll just merge this in and likely follow up with cleaning this up another time.

PR #58, opened 2016-12-15
langserver: Correctly detect no GlobalTracer

keegancsmith:
`opentracing.GlobalTracer()` by default is non-nil since it is the NoopTracer.
This means currently we will never use `telemetry/event`, even if we want to.

### DIFF
--- a/langserver/tracing.go
+++ b/langserver/tracing.go
@@ -23,7 +23,8 @@ func (h *HandlerCommon) InitTracer(conn *jsonrpc2.Conn) {
 		return
 	}
 
-	if hasGlobalTracer() {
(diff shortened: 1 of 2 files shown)


langserver/tracing.go
```diff
@@ -86,6 +84,17 @@ func (h *HandlerCommon) SpanForRequest(ctx context.Context, buildOrLang string,
 	return span, opentracing.ContextWithSpan(ctx, span), nil
 }
 
+// hasGlobalTracer checks if someone has setup a global tracer for this
+// process.
+func hasGlobalTracer() bool {
+	t := opentracing.GlobalTracer()
+	if t == nil {
```
neelance: `GlobalTracer()` can never be `nil`. That would break so many code paths.

PR #60, opened 2016-12-16
langserver: ctxvfs.NameSpace that is safe for concurrent read/writes

keegancsmith:
ctxvfs.NameSpace is not safe to use for concurrent read/writes (were a write is
a call to .Bind). As such we create a copy every time we want to use it in a
build.Context + it actually had some race conditions outside of that code. As
part of the perf work I am doing, I want to be able to bind to the NameSpace and
have it instantly appear to all BuildContexts. So this commit introduces
AtomicFS, a wrapper around NameSpace optimized for reads (since Bind calls are
relatively rare). The implementation avoids creating a copy every time we want a
BuildContext (common operation in our build server) + is completely safe for
concurrent access.

### DIFF
--- a/langserver/build_context.go
+++ b/langserver/build_context.go
@@ -36,17 +36,9 @@ func (h *LangHandler) defaultBuildContext() *build.Context {
 }
 
 func (h *HandlerShared) OverlayBuildContext(ctx context.Context, orig *build.Context, useOSFileSystem bool) *build.Context {
-	// Copy to avoid a race condition if concurrent goroutines
-	// bind new mountpoints to the ctxvfs.NameSpace. This lets us lock
-	// once here, instead of once per call to any ctxvfs.FileSystem
-	// method.
 	h.Mu.Lock()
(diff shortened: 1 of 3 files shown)


langserver/build_context.go
```diff
@@ -36,17 +36,9 @@ func (h *LangHandler) defaultBuildContext() *build.Context {
...
-	// bind new mountpoints to the ctxvfs.NameSpace. This lets us lock
-	// once here, instead of once per call to any ctxvfs.FileSystem
-	// method.
 	h.Mu.Lock()
-	fs := make(ctxvfs.NameSpace, len(h.FS))
-	for path, mounts := range h.FS {
-		fs[path] = mounts
-	}
+	fs := h.FS
 	h.Mu.Unlock()
```
keegancsmith: we still keep this mutex due to potential race conditions to the pointer h.FS

at a later stage I'll clean up all this ceremony around needing to protect h's fields due to calls to `Reset`. But that is unrelated to this PR

PR #62, opened 2016-12-19
Signature help provider

### LINKED ISSUE #75: Support textDocument/signatureHelp

alexsaveliev:
Experimenting with signature help provider feature. It works but pretty slow for now when comparing with vscode-go (uses `godef`)

Recording this PR to save my changes just in case

Fixes #75

### DIFF
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -184,6 +184,7 @@ func (h *LangHandler) Handle(ctx context.Context, conn JSONRPC2Conn, req *jsonrp
 				WorkspaceSymbolProvider:      true,
 				XWorkspaceReferencesProvider: true,
 				XDefinitionProvider:          true,
+				SignatureHelpProvider:        &lsp.SignatureHelpOptions{TriggerCharacters: []string{"(", ","}},
 			},
 		}, nil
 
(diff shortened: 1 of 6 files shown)


keegancsmith:
So this may be a bit slow due to how we invalidate the typecheck cache (each keystrokes throws away all the typechecking we do). However, this is actually a feature we want to support! So thanks :)

JeanMertz:
> So this may be a bit slow due to how we invalidate the typecheck cache

It seems this slowness is still an issue, am I right?

I ran into this with Sublime Text 3, as described in this issue:

https://github.com/tomv564/LSP/issues/318

Any thoughts on how this specific call can be sped up?

JeanMertz:
Also one more addition, it seems that the returned parameters are different between the `hover` response and the `signatureHelp` response:

screen shot 2018-05-06 at 15 57 45

screen shot 2018-05-06 at 16 19 39

The first one actually shows the package name if required (`zapcore`), the second doesn't.

keegancsmith:
Yeah. The difference you are noticing between hover and signature helper is due to us switching our implementation of hover to being based on godef (ie much faster). Signature helper is still using our type cache.

This is still an issue. However, there is some Go upstream work around making the libraries we use faster via caching. So for now we aren't tackling this, but when `go/types`, etc supports caching expect the go-langserver to be much faster.

PR #65, opened 2016-12-24
Query.CaseSensitive:

mbana:
* add flag to indicate if Query is CaseSensitive.
* add tests for, the parsed Query string and the expected output.


.gitignore
```diff
@@ -0,0 +1,44 @@
...
+node_modules
+jspm_packages
+
+# Optional npm cache directory
+.npm
+
+# Optional REPL history
+.node_repl_history
+
+.dev-tmp/
```
keegancsmith: did you accidentally add this? Seems to contain a bunch of git ignore which doesn't affect us.

felixfbecker:
Is this speced anywhere?

.gitignore
mbana: yes, soz.

slimsag:
The case sensitive portion of this change, AFAIK, fixes a real bug. so that part LGTM. The tests I'm also OK with. Let's rebase and land this.

@felixfbecker there is no spec for this because it's all just the Go LS interpretation of the spec.

@keegancsmith it's true we don't need a lot of the query stuff anymore; so we can remove & simplify a lot of it.

langserver/symbol.go
```diff
@@ -66,14 +71,44 @@ func queryJoin(s, e string) string {
 	return strings.TrimSpace(s + " " + e)
 }
 
+func parseCaseSensitive(q string) (qu Query, fields []string) {
+	reflectVal := reflect.Indirect(reflect.ValueOf(qu)).Type()
+	FieldIndex := reflectVal.NumField() - 1
+	Field := reflectVal.Field(FieldIndex)
+	// Field: %+v: {Name:CaseSensitve PkgPath: Type:bool Tag: Offset:80 Index:[5] Anonymous:false}
+	FieldName := Field.Name
+	FieldNamePrefix := fmt.Sprintf("%s:", FieldName)
```
keegancsmith: why are you using reflection here??

langserver/symbol.go
```diff
@@ -57,6 +61,7 @@ func (q Query) String() string {
 	for _, token := range q.Tokens {
 		s = queryJoin(s, token)
 	}
+	s = queryJoin(s, fmt.Sprintf("CaseSensitve:%v", q.CaseSensitve))
```
keegancsmith: typo

langserver/symbol.go
mbana: hi,
i was just thinking if it were possible to have codegen tool parse to `Query` instead of writing the code to this...
a cursory look makes me think that it might be possible?

i'll get rid of the reflection.

langserver/symbol.go
mbana: ta

mbana:
i thought github has an option to rebase somewhere?

mbana:
updated pr: https://github.com/sourcegraph/go-langserver/pull/105/files.

keegancsmith:
Closing this one out for the new one.

PR #68, opened 2017-01-09
optimize workspace/xreferences (use hints)

slimsag:
See https://github.com/sourcegraph/language-server-protocol/pull/10


beyang:
Add a test case.

Otherwise, LGTM

PR #82, opened 2017-01-12
langserver: Faster package list

### LINKED ISSUE #70: vscode: Support Workspace Symbol Search
Our symbol search currently does not seem to work when used via vscode. It only works in production sourcegraph.com due to some extra translations our build server does. It should work in vscode.

keegancsmith:
`buildutil.ExpandPattern` would inspect every package in the `GOPATH` and `GOROOT`.
The slowness of this is not much of a factor on sourcegraph.com, but for local
`GOPATH`s which can be large (such as my own), it is really slow. This takes
symbol search on this project on my machine from half a minute to instant.

This code is essentially the `buildutil.ExpandPattern` modified for our usecase. The idea
used can probably be submitted upstream since there are TODOs hinting at wanting this
functionality.

Previously we also incorrectly relied on `h.init.RootImportPath`, which is
technically not allowed in the langserver. In the future, we should work out
ways to prevent this accidental use of it.

Test Plan: Tested on the extensive integration tests for sourcegraph.com, which
will test both golang/go and other popular pkgs.

Fixes #70 and #69

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -278,8 +278,7 @@ func (h *LangHandler) handleSymbol(ctx context.Context, conn JSONRPC2Conn, req *
 		bctx := h.OverlayBuildContext(ctx, h.defaultBuildContext(), !h.init.NoOSFileSystemAccess)
 
 		par := parallel.NewRun(8)
-		pkgs := listPkgsUnderDir(bctx, rootPath)
-		for _, pkg := range pkgs {
+		for _, pkg := range listPkgsUnderDir(bctx, rootPath) {
 			// If we're restricting results to a single file or dir, ensure the
 			// package dir matches to avoid doing unnecessary work.
 			if results.Query.File != "" {
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -278,7 +278,7 @@ func (h *LangHandler) handleSymbol(ctx context.Context, conn JSONRPC2Conn, req *
 		bctx := h.OverlayBuildContext(ctx, h.defaultBuildContext(), !h.init.NoOSFileSystemAccess)
 
 		par := parallel.NewRun(8)
-		pkgs := listPkgs(bctx, rootPath)
+		pkgs := listPkgsUnderDir(bctx, rootPath)
 		for _, pkg := range pkgs {
 			// If we're restricting results to a single file or dir, ensure the
 			// package dir matches to avoid doing unnecessary work.
@@ -480,7 +480,13 @@ func init() {
 	prometheus.MustRegister(symbolCacheTotal)
 }
 
-func listPkgs(ctxt *build.Context, dir string) []string {
+// listPkgsUnderDir is buildutil.ExpandPattern(ctxt, []string{dir +
+// "/..."}). The implementation is modified from the upstream
+// buildutil.ExpandPattern so we can be much faster. buildutil.ExpandPattern
+// looks at all directories under GOPATH if there is a `...` pattern. This
+// instead only explores the directories under dir. In future
+// buildutil.ExpandPattern may be more performant (there are TODOs for it).
+func listPkgsUnderDir(ctxt *build.Context, dir string) []string {
 	ch := make(chan string)
 
 	var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


langserver/symbol.go
```diff
@@ -488,3 +479,91 @@ var symbolCacheTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
 func init() {
 	prometheus.MustRegister(symbolCacheTotal)
 }
+
+func listPkgs(ctxt *build.Context, dir string) []string {
```
slimsag: Perhaps rename to `expandPattern`, and document with what you put in your PR description:

> This code is essentially the buildutil.ExpandPattern modified for our usecase. The idea
used can probably be submitted upstream since there are TODOs hinting at wanting this
functionality.

Also, there are other callers of `buildutil.ExpandPatterns` in the langserver, like in `workspace_refs.go` IIRC. If you don't want to tackle that now -- that's OK, just throwing it out there in case you missed it :)

langserver/symbol.go
```diff
@@ -488,3 +479,91 @@ var symbolCacheTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
...
+	start = filepath.Clean(start) + string(os.PathSeparator)
+
+	if strings.HasPrefix(root, start) {
+		// If we are a child of start, we can just start at the
+		// root. A concrete example of this happening is when
+		// root=/goroot/src and start=/goroot
+		start = root
+	}
+
+	if !strings.HasPrefix(start, root) {
```
slimsag: `PathHasPrefix`, defined in `langserver/util.go`, is a more fool-proof way to do this.

langserver/symbol.go
```diff
@@ -488,3 +479,91 @@ var symbolCacheTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
...
+			return
+		}
+
+		pkg := filepath.ToSlash(strings.TrimPrefix(dir, root))
+
+		// Prune search if we encounter any of these import paths.
+		switch pkg {
+		case "builtin":
+			return
+		}
```
slimsag: Will there ever be other import paths here? If not, maybe just use `if`?

langserver/symbol.go
```diff
@@ -488,3 +479,91 @@ var symbolCacheTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
 func init() {
 	prometheus.MustRegister(symbolCacheTotal)
 }
+
+func listPkgs(ctxt *build.Context, dir string) []string {
+	ch := make(chan string)
```
slimsag: I wonder if making this channel buffered would help perf?

slimsag:
Along the lines of https://github.com/sourcegraph/go-langserver/issues/85 -- maybe this could also go into a subpackage somewhere. Just a quick thought -- not saying this is a perfect candidate or anything.

langserver/symbol.go
keegancsmith: This code is not at all `ExpandPattern`. I'll give it a better name and documentation though and mention why it is better than `ExpandPattern` and how it relates.

Yeah I'll avoid tackling other expand pattern uses, but I'll add an inline TODO to hint to future travels about perf improvements.

langserver/symbol.go
keegancsmith: `strings.HasPrefix` is the same technique used in the upstream code I based this off. Notice the appending of `os.Seperator`. However, I'll take a look at `PathHasPrefix` and likely switch to it.

langserver/symbol.go
keegancsmith: This is verbatim from the upstream code. I'd rather leave it as is to make potential diffing we do in future clearer.

langserver/symbol.go
keegancsmith: I would expect it would potentially make perf worse! The thing consuming this channel is a very tight loop appending to a slice. So will be like a buffered channel, without the need of extra ceremony that buffered channel likely needs to do.

langserver/symbol.go
keegancsmith: The behaviour of PathHasPrefix is slightly different to what we are doing here. Given we already need the cleaned up root and start above, I think this code is clearer since when reading it we know exactly what strings.HasPrefix does.

langserver/symbol.go
keegancsmith: I updated workspace_refs as well, since it was straightforward.

PR #84, opened 2017-01-13
langserver: workspace/xreferences: change Go DependencyReference.Hints dir -> dirs

slimsag:
This will give us much less DB entries (one/dependency referenced/repo vs. one/dependency
referenced/directory/repo).

Additionally, this will help with optimizing `workspace/xreferences` to be both performant
and accurate. It will however mean that we will likely need to add a `Limit` or other
pagination options to `workspace/xreferences`.

Also update the tests accordingly.


slimsag:
@beyang adding you as a reviewer because this change is small, uncontroversial, and I'd like to get it out asap for reindexing.

PR #86, opened 2017-01-13
Windows support

alexsaveliev:
- dealing with Windows-specific paths, converting them to Unix style where needed.

### DIFF
--- a/langserver/ast.go
+++ b/langserver/ast.go
@@ -59,7 +59,7 @@ func goRangesToLSPLocations(fset *token.FileSet, nodes []*ast.Ident) []lsp.Locat
 
 func goRangeToLSPLocation(fset *token.FileSet, pos token.Pos, end token.Pos) lsp.Location {
 	return lsp.Location{
-		URI:   "file://" + fset.Position(pos).Filename,
+		URI:   pathToURI(fset.Position(pos).Filename),
 		Range: rangeForNode(fset, fakeNode{p: pos, e: end}),
 	}
 
--- a/langserver/build_context.go
+++ b/langserver/build_context.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"runtime"
 	"strings"
 
 	"github.com/sourcegraph/ctxvfs"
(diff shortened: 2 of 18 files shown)


langserver/ast.go
```diff
@@ -59,7 +59,7 @@ func goRangesToLSPLocations(fset *token.FileSet, nodes []*ast.Ident) []lsp.Locat
 
 func goRangeToLSPLocation(fset *token.FileSet, pos token.Pos, end token.Pos) lsp.Location {
 	return lsp.Location{
-		URI:   "file://" + fset.Position(pos).Filename,
```
keegancsmith: why can't we return file URIs like we did before? What makes this incorrect on windows?

alexsaveliev: On Windows, file URL uses triple slash, see https://en.wikipedia.org/wiki/File_URI_scheme#Windows
I'm trying to use the same function everywhere

keegancsmith: We also return triple slash on unix, since `fset.Position(pos).Filename` starts with `/`. I'm surprised we can't do this on windows.

How easy is it to get a windows environment I can SSH or RDP into? This change touches a lot of places and is the sort of thing that seems like it would break in future. I'd like to confirm it is necessary.

Relatedly, can't we change `vfs.OS` so it returns paths that just work?

alexsaveliev: > We also return triple slash on unix, since fset.Position(pos).Filename starts with /. I'm surprised we can't do this on windows.

On Windows, filename will be `C:/foo/bar` thus making `file://C:/foo/bar` while we need `file:///C:/foo/bar`
Also, I believe that refactoring to introduce common path2uri function is good as is :)

> How easy is it to get a windows environment I can SSH or RDP into? 

I'm using neat feature Appveyor https://www.appveyor.com/docs/how-to/rdp-to-build-worker/ to connect to build server via RDP, you have 1 hour to experiments. Hope this helps

> Relatedly, can't we change vfs.OS so it returns paths that just work?
Won't be enough. `vfs.OS` may return list of directory entries, for example, and then `build.Context` will be responsible for constructing path elements...

felixfbecker: `path2uri` implementations typically need to add a leading slash if the path matches a windows path. That is typically done by testing against a drive letter followed by a colon as the first path segment.

langserver/build_context.go
```diff
@@ -42,35 +41,29 @@ func (h *HandlerShared) OverlayBuildContext(ctx context.Context, orig *build.Con
 	ctxt := &copy
 
 	ctxt.OpenFile = func(path string) (io.ReadCloser, error) {
-		return fs.Open(ctx, path)
+		return fs.Open(ctx, vfsPath(path))
```
keegancsmith: why does this wrapper need to exist? Shouldn't users of build context only be using virtual paths?

langserver/build_context_test.go
```diff
@@ -13,11 +13,12 @@ func TestContainingPackage(t *testing.T) {
...
 		},
 	})
+
 	bctx.GOPATH = "/go"
 
 	tests := map[string]string{
-		"/go/src/p/a.go":      "p",
-		"/go/src/p/a_test.go": "p_test",
+		bctx.GOPATH + "/src/p/a.go":      "p",
+		bctx.GOPATH + "/src/p/a_test.go": "p_test",
```
keegancsmith: why does this make a difference? Line 17 sets bctx.GOPATH to `"/go"`

alexsaveliev:
closing in favor of #133

PR #87, opened 2017-01-16
langserver: Cleanup BuildContext functions

keegancsmith:
This commit contains a few fixes to how we export and use our Build Context
related functions. We can greatly simplify this due to changes like AtomicFS and
changes on sourcegraph.com's build server side.

We used to export OverlayBuildContext since at one point we would use it before
doing init on sourcegraph.com's build server. Now we use the same build context
after init, so we can combine both defaultBuildContext and OverlayBuildContext
and use that everywhere. This will require a change to sourcegraph.com's
codebase.

### DIFF
--- a/langserver/build_context.go
+++ b/langserver/build_context.go
@@ -1,7 +1,6 @@
 package langserver
 
 import (
-	"context"
 	"fmt"
 	"go/build"
 	"io"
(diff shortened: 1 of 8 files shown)


langserver/build_context.go
```diff
@@ -10,12 +9,15 @@ import (
...
 	"strings"
+
+	"golang.org/x/net/context"
 )
 
-func (h *LangHandler) defaultBuildContext() *build.Context {
-	bctx := &build.Default
+// BuildContext creates a build.Context which uses the overlay FS and the InitializeParams.BuildContext overrides.
+func (h *LangHandler) BuildContext(ctx context.Context) *build.Context {
+	var ctxt *build.Context
```
slimsag: nit: doesn't the old `bctx` var name make more sense given `ctx` is already in scope here from the parameter?

keegancsmith: Sure, I'll update.

PR #88, opened 2017-01-16
langserver: Use h.FindPackage for refs loader.Config

keegancsmith:
This allows us to also use the same FindPackage used by our typechecker.

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -47,14 +47,15 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
 
 	// Perform typechecking.
 	var (
-		fset = token.NewFileSet()
-		pkgs []string
+		findPackage = h.getFindPackageFunc()
+		fset        = token.NewFileSet()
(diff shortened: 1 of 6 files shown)


keegancsmith:
@slimsag there are other uses of `Import` which could be updated. I'm not sure if they would benefit from using the custom FindPackage (ie like JIT dependency fetching/etc). Do you mind taking a look at them and commenting if they would need all dependencies fetched. Most of the uses seem to be related to finding a package under rootPath, but I could be wrong.

langserver/workspace_refs.go
```diff
@@ -130,6 +130,11 @@ func (h *LangHandler) workspaceRefsTypecheck(ctx context.Context, bctx *build.Co
 		span.Finish()
 	}()
 
+	findPackage := h.FindPackage
+	if findPackage == nil {
+		findPackage = defaultFindPackageFunc
+	}
```
slimsag: @keegancsmith sorry I should've been more clear on Slack, I was preparing a PR that does exactly this among other things :)

PR #89, opened 2017-01-16
workspace/xreferences: optimizations; correct vendor handling; minor cleanups

slimsag:
Based on #88 (I will drop those commits once it is merged)

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -3,6 +3,7 @@ package langserver
 import (
 	"context"
 	"fmt"
+	"go/ast"
 	"go/build"
 	"go/parser"
 	"go/token"
@@ -13,10 +14,10 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"golang.org/x/tools/go/loader"
 
-	"github.com/neelance/parallel"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/sourcegraph/go-langserver/langserver/internal/refs"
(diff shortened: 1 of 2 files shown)


keegancsmith:
Rebased for you, will review now.

langserver/workspace_refs.go
```diff
@@ -70,27 +72,36 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
...
-	// Collect dependency references.
-	results := refResultSorter{results: make([]lspext.ReferenceInformation, 0)}
-	par := parallel.NewRun(parallelism)
-	for _, pkg := range prog.Imported {
-		par.Acquire()
-		go func(pkg *loader.PackageInfo) {
-			defer par.Release()
+	// Collect dependency references in the AfterTypeCheck phase. This enables
+	// us to begin looking at packages as they are typechecked, instead of
+	// waiting for all packages to be typechecked (which is IO bound).
```
keegancsmith: neat

langserver/workspace_refs.go
```diff
@@ -70,27 +72,36 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
...
+		started      uint32
+	)
+	afterTypeCheck := func(pkg *loader.PackageInfo, files []*ast.File) {
+		_, interested := unvendoredPackages[pkg.Pkg.Path()]
+		if !interested {
+			return
+		}
+
+		// Do not block the type-checker, but also avoid running with unbounded
+		// parallelism.
```
keegancsmith: given typechecking is IO bound, surely this function is faster? Can't we just do unbounded parallelism to simplify?

langserver/workspace_refs.go
```diff
@@ -100,19 +111,29 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
...
+		}()
+	}
+
+	_, err := h.workspaceRefsTypecheck(ctx, bctx, conn, fset, pkgs, afterTypeCheck)
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait for all worker goroutines to complete.
+	for i := 0; i < int(started); i++ {
```
keegancsmith: technically you should read started via the atomic pkg. I do realise though that when we get here, there should be no concurrent accesses going on.

Also it seems your started + finishSignal could be accomplished with just a `sync.WaitGroup`

slimsag:
Thanks for the rebase & review! I've used the `sync.WaitGroup` as you suggested, I knew I was missing something there (seemed oddly difficult to do).

PR #91, opened 2017-01-17
symbol: single-flight cache access

keegancsmith:
See commit messages

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -336,17 +336,18 @@ type pkgSymResult struct {
 func (h *LangHandler) collectFromPkg(ctx context.Context, bctx *build.Context, pkg string, rootPath string, results *resultSorter) {
 	h.pkgSymCacheMu.Lock()
 	res, ok := h.pkgSymCache[pkg]
-	if !ok {
-		res = &pkgSymResult{ready: make(chan struct{})}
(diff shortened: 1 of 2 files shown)


langserver/symbol.go
```diff
@@ -325,14 +325,28 @@ func (h *LangHandler) handleSymbol(ctx context.Context, conn JSONRPC2Conn, req *
...
+	res, ok := h.pkgSymCache[pkg]
+	if !ok {
+		res = &pkgSymResult{ready: make(chan struct{})}
+		h.pkgSymCache[pkg] = res
+		defer close(res.ready)
+	}
 	h.pkgSymCacheMu.Unlock()
-	if pkgSyms == nil {
+
+	if ok {
```
neelance: I think this would be much easier to read if you combine this `if` with the `if` above. Also I would suggest to use a second function and do an early-exit on cache hit.

keegancsmith: I'm not sure if the 2nd suggestion will improve it that much, especially since the function signature is quite long. Sent an update for just the first, but can do the 2nd if you think it is important.

What I liked about how this was before was it was super obvious which parts of code held the `pkgSymCacheMu` mutex since we only unlocked it in one place.

PR #92, opened 2017-01-17
langserver: Make caches bounded at the process level

keegancsmith:
We experience OOMs on a semi-regular basis at the moment in sourcegraph production. This is due to our caches being unbounded. This PR is an experiment at bounding the caches. We bound them at the process level since in sourcegraph.com production we can have multiple active langserver in the same process. The initial cache sizes are estimates at what a useful size is. After experimentation with this having production traffic, I'll update the defaults.

### DIFF
--- a/langserver/cache.go
+++ b/langserver/cache.go
@@ -2,8 +2,23 @@ package langserver
 
 import (
 	"sync"
+	"sync/atomic"
 
 	lru "github.com/hashicorp/golang-lru"
+)
+
+var (
+	// typecheckCache is a process level cache for storing typechecked
+	// values. Do not directly use this, instead use newTypecheckCache()
+	typecheckCache = newARC(1000)
+
+	// symbolCache is a process level cache for storing symbols found. Do
+	// not directly use this, instead use newSymbolCache()
+	symbolCache = newARC(1000)
+
+	// cacheID is used to prevent key conflicts between different
+	// LangHandlers in the same process.
+	cacheID int64
 )
 
 type cache interface {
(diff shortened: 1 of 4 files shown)


langserver/cache.go
```diff
@@ -0,0 +1,132 @@
...
+	// the cache.
+	c.mu.Lock()
+	c.id = nextCacheID()
+	c.mu.Unlock()
+}
+
+// newARC is a wrapper around lru.NewARC which does not return an error.
+func newARC(env string, defaultSize int) *lru.ARCCache {
+	size := defaultSize
+	if i, err := strconv.Atoi(os.Getenv(env)); err == nil && i > 0 {
```
neelance: Could we please use the `pkg/env` package here instead of `os.Getenv`?

langserver/handler.go
```diff
@@ -71,19 +67,24 @@ func (h *LangHandler) resetCaches(lock bool) {
...
+
 	h.importGraphOnce = sync.Once{}
 	h.importGraph = nil
-	if lock {
-		h.mu.Unlock()
+
+	if h.typecheckCache == nil {
+		h.typecheckCache = newTypecheckCache()
+	} else {
+		h.typecheckCache.Purge()
```
neelance: Why do we need `Purge()` here? Looks to me as if `newTypecheckCache()` would be nearly identical.

langserver/cache.go
keegancsmith: `pkg/env` doesn't exist in this repo. But can vendor it in.

langserver/handler.go
keegancsmith: You are correct, will update. Initially I thought I'd invalidate items from the cache, but there is no need for that.

langserver/cache.go
neelance: Ah right. Then it's fine. Whatever you prefer.

langserver/handler.go
keegancsmith: Purge exists due to the mutex for the variable. If we just do `h.typecheckCache = newTypecheckCache()` we have a race condition on setting `h.typecheckCache`.

I have future plans to fix the need to protect setting these variables.

neelance: Hmm, but this is a bit weird that `h.typecheckCache == nil` determines if `h.typecheckCache` can have a race condition or not. I'll approve this PR now, I trust you to keep this in mind. :-)

keegancsmith: yes it is weird!! this whole bit of the code is one of the things I was aiming to cleanup this week. Basically we rely on resetCaches being called when initialize request is sent. Before initialize is sent everything is nil and we enforce initialize is the first request. Before initialize is done, every other request is either blocked or rejected. Then this bit of code is used again for clearing caches when the user edits a file (not used in production, but used in desktop environments).

Its all pretty crappy TBH. The major change I wanted to make in arch was separating out the init from the reset, so we don't have think about protecting assignment of every variable in LangHandler.

PR #94, opened 2017-01-18
implement workspace/symbol extension

slimsag:
This PR implements the extended workspace/symbol method, per the spec @ https://github.com/sourcegraph/language-server-protocol/blob/master/extension-workspace-references.md#extended-workspace-symbol-request

Additionally, it adds an `id` field to our `SymbolDescriptor` to support Sourcegraph URL-To-Definition links. See the commit message https://github.com/sourcegraph/go-langserver/commit/1dcf5891d33c375b5bf9ada073214b3b108bbb80 for details.

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -269,13 +269,60 @@ func toSym(name string, bpkg *build.Package, recv string, kind lsp.SymbolKind, f
 			"packageName": bpkg.Name,
 			"recv":        recv,
 			"name":        name,
+			"id":          fmt.Sprintf("%s:%s:%s:%s", path.Clean(bpkg.ImportPath), bpkg.Name, recv, name),
 		},
 	}
 }
 
+func parseID(v string) (pkg, packageName, recv, name string) {
+	f := strings.Split(v, ":")
+	if len(f) != 3 && len(f) != 4 {
+		log.Printf("parseID: invalid ID encountered: %q", v)
+		return "", "", "", ""
+	}
+	if len(f) == 3 {
+		return f[0], "", f[1], f[2]
+	}
+	return f[0], f[1], f[2], f[3]
+}
+
+// idEqual tells if two SymbolDescriptor IDs are equal. The Go IDs are in the form:
+//
+//  <package>:<packageName>:<recv>:<name>
+//
+// Additionally, an extra form with <packageName> omitted entirely is supported:
+//
+//  <package>:<recv>:<name>
+//
+// Examples:
+//
+//  github.com/gorilla/mux:mux_test:Router:ServeHTTP
+//  github.com/gorilla/mux:Router:ServeHTTP
+//
+// When <packageName> is omitted in either a or b, idEqual simply makes no
+// comparison of that field. That is,t he above two examples are said to be
+// equal.
+func idEqual(a, b string) bool {
+	aPkg, aPackageName, aRecv, aName := parseID(a)
+	bPkg, bPackageName, bRecv, bName := parseID(b)
+	if aPackageName == "" || bPackageName == "" {
+		return aPkg == bPkg && aRecv == bRecv && aName == bName
+	}
+	return aPkg == bPkg && aPackageName == bPackageName && aRecv == bRecv && aName == bName
+}
+
 // symbolContains tells if a exactly contains b.
+//
+// The only exception is the `id` field, which is treated specially. Instead of
+// a byte-by-byte string comparison, idEqual is used.
 func symbolContains(a, b lspext.SymbolDescriptor) bool {
 	for k, v := range a {
+		if k == "id" {
+			if bv, ok := b[k]; !ok || !idEqual(v.(string), bv.(string)) {
+				return false
+			}
+			continue
+		}
 		if bv, ok := b[k]; !ok || bv != v {
 			return false
 		}
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/neelance/parallel"
 	"github.com/sourcegraph/go-langserver/pkg/lsp"
+	"github.com/sourcegraph/go-langserver/pkg/lspext"
 	"github.com/sourcegraph/jsonrpc2"
 )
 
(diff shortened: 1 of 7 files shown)


langserver/symbol.go
```diff
@@ -239,17 +250,84 @@ func score(q Query, s lsp.SymbolInformation) (scor int) {
...
+			"package":     path.Clean(bpkg.ImportPath),
+			"packageName": bpkg.Name,
+			"recv":        recv,
+			"name":        name,
+			"id":          fmt.Sprintf("%s:%s:%s:%s", path.Clean(bpkg.ImportPath), bpkg.Name, recv, name),
+		},
+	}
+}
+
+func parseID(v string) (pkg, packageName, recv, name string) {
```
beyang: This is exactly the kind of thing we want to avoid with an `id` field. The IDs should only be compared on the basis of string equality. If we start encoding structure into the ID, we're basically trying to encode a `SymbolDescriptor` into a string format. In that case, we might as well use a standard encoding (like URL encoding). This also means we're treating `id` differently from other `SymbolDescriptor` fields, which is hacky and unexpected.

pkg/lsp/service.go
```diff
@@ -62,6 +62,11 @@ type ServerCapabilities struct {
 	// XDefinitionProvider indicates the server provides support for
 	// textDocument/xdefinition. This is a Sourcegraph extension.
 	XDefinitionProvider bool `json:"xdefinitionProvider,omitempty"`
+
+	// XWorkspaceSymbolByProperties indicates the server provides support for
+	// querying symbols by properties with WorkspaceSymbolParams.symbol. This
+	// is a Sourcegraph extension.
+	XWorkspaceSymbolByProperties bool `json:"xworkspaceSymbolByProperties,omitempty"`
```
beyang: Why not call this `XWorkspaceSymbolProvider`, since it is intended as a replacement for `WorkspaceSymbolProvider`?

slimsag: This is just following the name in the spec. @felixfbecker thoughts?

felixfbecker: @beyang because that would imply that `workspace/symbol` is a new method. What's new is only the ability to query by symbol properties.

langserver/symbol.go
felixfbecker: So here is the use case we need to solve - at the moment for Go, soon for as many languages as possible:
- We **have** a unique, human-readable symbol name (referred to as symbol ID)
- We **need** symbol information about the symbol it refers to (location to be precise)

To be unique, the symbol ID will be composed of different parts, in **all languages**.
What parts exactly need to be encoded inside the string and how they are separated the best is **language-specific**.

So, somewhere in the the whole process that solves above use case, the symbol ID needs to be decomposed into its language-specific parts. This requires language-specific code. There are two options:

1. Let the frontend parse the symbol ID (from the URL) in a huge `switch` with language-specific patterns and build a `SymbolDescriptor` out of the part, then query by that.
2. Let the frontend just query by the symbol ID directly, e.g. pass it directly from the URL into a `SymbolDescriptor` that looks like `{ id: idFromUrl }` and query by that. The _language server_ will parse the ID and in the language-specific way. _The language-specific code will stay inside the language servers._

We want 2) because we want to _avoid_ language-specific code outside of language servers, because they should abstract away language specificness by speaking the common protocol that is LSP.

Does that make sense?

felixfbecker:
@slimsag Can you comment on the pattern you decided on for the Go symbol IDs?

slimsag:
@felixfbecker I will msg you on slack, but also see the commit msg I linked to in the description

langserver/symbol.go
```diff
@@ -239,17 +250,41 @@ func score(q Query, s lsp.SymbolInformation) (scor int) {
...
-		Name:          name,
-		Kind:          kind,
-		Location:      goRangeToLSPLocation(fs, pos, pos+token.Pos(len(name))-1),
-		ContainerName: container,
+}
+
+// symbolContains tells if a exactly contains b.
+//
+// The only exception is the `id` field, which is treated specially. Instead of
+// a byte-by-byte string comparison, idEqual is used.
```
keegancsmith: @slimsag you don't specially treat the ID field. Also this implementation seems to tell you if b contains a, not a contains b.

On top of this, lspext.SymbolDescriptor seems to be quite the memory inefficient way to store this information. I am working on a fix since our symbol caches are pretty poor, and now we also store this as part of our symbol cache.

langserver/symbol.go
```diff
@@ -239,17 +250,41 @@ func score(q Query, s lsp.SymbolInformation) (scor int) {
...
+func symbolContains(a, b lspext.SymbolDescriptor) bool {
+	for k, v := range a {
+		if bv, ok := b[k]; !ok || bv != v {
```
keegancsmith: don't you need to use reflect.DeepEqual here since they are interface values?

langserver/symbol.go
keegancsmith: implemented memory efficient symbol descriptor at https://github.com/sourcegraph/go-langserver/pull/98

langserver/symbol.go
felixfbecker: Yes, the query should be matched with deep equality. I should also add that to the spec.

keegancsmith: @felixfbecker yeah I was actually commenting on interface equality. I thought golang would actually just compare interfaces objects (ie it would be like a pointer), but comparison operators on interfaces actually do something like
```
return a.Type == b.Type && a.ActualValue == b.ActualValue
```
TIL

felixfbecker: Deeply or shallow though?

keegancsmith: depends if your underlying types have pointers. In the case of this code, we don't have to worry about that.

felixfbecker: `SymbolDescriptor` can have `package`, which itself has properties. So my question is would it compare only the values of the properties that are present in `WorkspaceSymbolParams.symbol.package` then or would it do an exact match of the two?

keegancsmith: this implementation does not. However, it is a private function only used by go-langserver. And the SymbolDescriptors emitted by go-langserver are flat. ie everything here https://github.com/sourcegraph/go-langserver/pull/94/files/0004bd11e47796b6580e067de985313c0bb982ea#diff-41a3fbc5d9dbf547085a3235ea31fd2aR266 is just a string value (excludign vendor which is a bool)

PR #97, opened 2017-01-19
Refactoring and Travis CI for forked repositories

alexsaveliev:
- refactoring: extracted uri2path and path2uri methods. I'd like to have them defined in the single place plus I hope to make them support VFS, client, and OS-specific URIs and path elements
- forcing Go import path for Travis build to `github.com/sourcegraph/go-langserver` to be able to run Travis builds on forked repositories


langserver/util.go
```diff
@@ -31,3 +32,44 @@ func pathEqual(a, b string) bool {
...
+}
+
+// pathToUri converts given path to file URI
+func pathToUri(path string) string {
+	prefix := "file://"
+	if !strings.HasPrefix(path, "/") {
+		// On Windows there should be triple slash
+		prefix += "/"
+	}
+	path = strings.Replace(path, "\\", "/", -1)
```
keegancsmith: this breaks unix. You can have a file or directory called `a\b`. Can we instead rely on which operating we are running on to do this conversion?

alexsaveliev: > Can we instead rely on which operating we are running on to do this conversion?

It's more complicated then.. On Windows we may encounter slashes when working with VFS and backslashes (local FS)

felixfbecker: We can detect a Windows path by checking for a drive letter with colon. Colon is reserved on Unix

keegancsmith: @alexsaveliev any updates?

alexsaveliev: @keegancsmith, @felixfbecker addressed your remarks. If path looks like Windows-one (second rune is `:`), converting backslashes to slashes before converting it to URI, otherwise leaving it as is, url-encoding backslashes inside.

langserver/handler_common.go
```diff
@@ -27,10 +26,10 @@ func (h *HandlerCommon) Reset(rootURI string) error {
 	if h.shutdown {
 		return errors.New("unable to reset a server that is shutting down")
 	}
-	if !strings.HasPrefix(rootURI, "file:///") {
+	if !isUri(rootURI) {
 		return fmt.Errorf("invalid root path %q: must be file:/// URI", rootURI)
 	}
-	h.RootFSPath = strings.TrimPrefix(rootURI, "file://") // retain leading slash
+	h.RootFSPath = uriToPath(rootURI) // retain leading slash
```
keegancsmith: this is a path for our VFS, not for the OS

langserver/util.go
```diff
@@ -35,6 +36,56 @@ func IsVendorDir(dir string) bool {
...
+		path := strings.TrimPrefix(uri, "file://")
+		// On Windows, VS Code sends "file:///c%3A/..."
+		if unescaped, err := url.QueryUnescape(path); err == nil {
+			path = unescaped
+		}
+		// checking if we have a Windows-style URL such as file:///C:/
+		// if so, removing leading slash
+		if len(path) > 2 && path[0] == '/' && path[2] == ':' {
+			path = path[1:]
+		}
```
keegancsmith: `/C:/` is a valid unix path.

langserver/fs.go
```diff
@@ -99,7 +99,7 @@ func (h *HandlerShared) readFile(ctx context.Context, uri string) ([]byte, error
 }
 
 func uriToOverlayPath(uri string) string {
-	return strings.TrimPrefix(uri, "file:///")
+	return strings.TrimPrefix(uriToPath(uri), "/")
```
keegancsmith: this is different behaviour than before. If my string is `/hi`, before it will stay `/hi`, but with change it will become `hi`.

langserver/handler_common.go
```diff
@@ -27,10 +26,10 @@ func (h *HandlerCommon) Reset(rootURI string) error {
 	if h.shutdown {
 		return errors.New("unable to reset a server that is shutting down")
 	}
-	if !strings.HasPrefix(rootURI, "file:///") {
+	if !isUri(rootURI) {
```
keegancsmith: in some places we trick triple forward slash, in other only two slashes. However, you have replaced all implementations with just `isUri` which checks triple forward slash.

Is the places we checked double forward slash just not as correct and this change is correct?

langserver/util.go
```diff
@@ -35,6 +35,21 @@ func IsVendorDir(dir string) bool {
 	return strings.HasPrefix(dir, "vendor/") || strings.Contains(dir, "/vendor/")
 }
 
+// isUri tells if s denotes an URI
+func isUri(s string) bool {
```
keegancsmith: can you call this `isURI`

langserver/util.go
```diff
@@ -35,6 +35,21 @@ func IsVendorDir(dir string) bool {
...
+	return strings.HasPrefix(s, "file:///")
+}
+
+// pathToUri converts given path to file URI
+func pathToUri(path string) string {
```
keegancsmith: Can you call this `pathToURI`. Can we also document that path has to be absolute?

langserver/fs.go
alexsaveliev: fixed

langserver/handler_common.go
alexsaveliev: there was a single place where we have checked for double slash, but according to comments it's safe to use the same triple-slash-check as we do in other places, because we need to test `URI or not`

```		// HACK: RootPath is not a URI, but historically we treated it
		// as such. Convert it to a file URI
		if !strings.HasPrefix(params.RootPath, "file://") {
			params.RootPath = "file://" + params.RootPath
		}
```

langserver/util.go
alexsaveliev: done

langserver/util.go
alexsaveliev: done

PR #99, opened 2017-01-20
symbol: Off-by-1 in range

### LINKED ISSUE #7: off by one error observed between workspace/symbol and textDocument/definition calls
On our code base, run textDocument/definition with the following args:
textDocument:"file:///Users/mattfs/gopath/src/sourcegraph.com/sourcegraph/sourcegraph/app/app.go"
position: line: 26, character: 39

I get back the following symbol:
uri: "file:///usr/local/go/src/net/http/server.go"
range: 76/5 -> 76/11

Call workspace symbols on our code base, with the query "Handler." The token we are looking for, in file:///usr/local/go/src/net/http/server.go, has this range:
76/5 -> 76/12

Still learning up on LSP, but the above seems like it could be an off by one error! The reason I'd like these to be equal is I'm calling deep_equal on all the ranges returned by workspace/symbol to try and find the one corresponding to textDocument/definition .

keegancsmith:
The LSP spec states the end of ranges are exclusive.

Fixes #7

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -259,7 +259,7 @@ func toSym(name string, bpkg *build.Package, recv string, kind lsp.SymbolKind, f
 		SymbolInformation: lsp.SymbolInformation{
 			Name:          name,
 			Kind:          kind,
-			Location:      goRangeToLSPLocation(fs, pos, pos+token.Pos(len(name))-1),
+			Location:      goRangeToLSPLocation(fs, pos, pos+token.Pos(len(name))),
 			ContainerName: container,
 		},
 		// NOTE: fields must be kept in sync with workspace_refs.go:defSymbolDescriptor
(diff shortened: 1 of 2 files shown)


keegancsmith:
@slimsag any reason this would break any of our indexes on sourcegraph.com?

langserver/symbol.go
```diff
@@ -259,7 +259,7 @@ func toSym(name string, bpkg *build.Package, recv string, kind lsp.SymbolKind, f
 		SymbolInformation: lsp.SymbolInformation{
 			Name:          name,
 			Kind:          kind,
-			Location:      goRangeToLSPLocation(fs, pos, pos+token.Pos(len(name))-1),
+			Location:      goRangeToLSPLocation(fs, pos, pos+token.Pos(len(name))),
```
slimsag: I am surprised that this doesn't require updating tests... any clue why that is?

slimsag:
It won't break our indexes (they don't have any concept of line/col numbers)

langserver/symbol.go
keegancsmith: Our tests for symbols only check range start.

PR #100, opened 2017-01-20
langserver: change SymbolDescriptor id format to use slashes

slimsag:


slimsag:
For details see https://github.com/sourcegraph/sourcegraph/issues/3233#issuecomment-274165937

keegancsmith:
Ok I see now you are using `/-/` to delim, now you don't lose uniqueness in the same way I was thinking.

PR #101, opened 2017-01-21
langserver: add 15s timeout to workspace/xreferences

slimsag:
This adds a 15s timeout to workspace/xreferences which prevents some long-running requests from causing any trouble. This can happen in rarer cases with very large codebases, etc.

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -13,6 +13,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	"golang.org/x/tools/go/loader"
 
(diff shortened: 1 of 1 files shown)


langserver/workspace_refs.go
```diff
@@ -149,7 +161,14 @@ func (h *LangHandler) workspaceRefsTypecheck(ctx context.Context, bctx *build.Co
 			}
 			return bpkg, nil
 		},
-		AfterTypeCheck: afterTypeCheck,
+		AfterTypeCheck: func(pkg *loader.PackageInfo, files []*ast.File) {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
```
keegancsmith: you can simplify your done checks. Just do
```
if ctx.Err() != nil {
  return
}
```

this applies to all your other uses as well.

langserver/workspace_refs.go
```diff
@@ -23,7 +24,18 @@ import (
...
+// the same identical timeout in client requests. It cannot be entirely
+// enforced by the go-langserver, because go/loader doesn't support context
+// timeouts during lengthy type-checking operations.
+const WorkspaceReferencesTimeout = 15 * time.Second
+
 func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lspext.WorkspaceReferencesParams) ([]referenceInformation, error) {
+	// TODO: Find a better way to pass this timeout down as part of the request
+	// via LSP itself.
+	ctx, cancel := context.WithTimeout(ctx, WorkspaceReferencesTimeout)
+	defer cancel()
```
keegancsmith: We should add support for https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#cancelRequest

I've already seen vscode send it to cancel symbol requests. This seems like the sort of thing that should be controlled by the caller. In the meantime though, this is sufficient.

Issue #25, opened 2016-11-07
Definition range


nicot:
Definition range should span the entire function body and declaration. The Go AST position info for the def only has the position of the name in it, so we'd have to dig a little deeper to get this to work right.

Start here: https://github.com/sourcegraph/go-langserver/blob/master/langserver/definition.go#L29. These lines can be removed once the entire function body is included: https://github.com/sourcegraph/go-langserver/blob/master/langserver/definition.go#L44-L48.

keegancsmith:
@uforic should this be part of the editor extension project https://github.com/sourcegraph/go-langserver/projects/1

uforic:
Cool - @nicot , mind talking about the use case? Couldn't tell from the LSP protocol.md if this was desired behavior, just `definition location of a symbol`, which seems more like just the name.

Maybe the entire range could be an additional field in the resolvedSymbol request? : https://github.com/sourcegraph/language-server-protocol/pull/2

nicot:
This is the way that VSCode expects it to work, so that we can show declarations in the hover tooltip. This is the way that most language servers work, so it should not be an extension.

uforic:
Cool - makes sense. Let's submit a PR to the LSP folks to this affect, the language on the protocol is kinda vague.

@nicot , can you link to the code where they expect the range to encompass the whole function? Will help with the PR.

nicot:
This is the gotoDeclaration command that the hover uses to provide a preview: https://github.com/Microsoft/vscode/blob/master/src/vs/editor/contrib/goToDeclaration/browser/goToDeclaration.ts#L329.

slimsag:
> Cool - makes sense. Let's submit a PR to the LSP folks to this affect, the language on the protocol is kinda vague.

@uforic I'm all for clarification of docs :D but tbh I think it's already straightforward:

1. `textDocument/definition` returns a `Location`: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#location
 
2. `Location` is defined as:

```
interface Location {
    uri: string;
    range: Range;
}
```

And to quote the `Range` docs:

> #### Range
> 
> A range in a text document expressed as (zero-based) start and end positions. A range is comparable to a selection in an editor. Therefore the end position is exclusive.
> 
> ```typescript
> interface Range {
> 	/**
> 	 * The range's start position.
> 	 */
> 	start: Position;
> 
> 	/**
> 	 * The range's end position.
> 	 */
> 	end: Position;
> }
[...]

PR #102, opened 2017-01-22
Send back the location of the identifier

nicot:
Addresses: https://github.com/sourcegraph/go-langserver/issues/25

Allows: https://github.com/sourcegraph/sourcegraph/issues/3322


keegancsmith:
cc @slimsag can this change cause any issues in the global refs pipeline?

PR #103, opened 2017-01-22
textDocument/xdefinition: do not return "nil" AND no error.

slimsag:
The return type signature of `handleDefinition` is `([]lsp.Location, error)`. If `nil, nil`
is returned then ` (*LangHandler).Handle` also returns `result interface{}` as `nil`. When
the result of a jsonrpc2 handler is nil, `jsonrpc2.HandlerWithError` [converts it into a struct](https://sourcegraph.com/github.com/sourcegraph/jsonrpc2/-/blob/handler_with_error.go#L25-28)
which is then encoded as a JSON object instead of an array. When decoding, this results in:

```
json: cannot unmarshal object into Go value of type []lspext.SymbolLocationInformation
```

Which is (obviously) bad because that looks like an error instead of 'no results'.

According to that jsonrpc2 implementation (linked above):

```
// isNilValue tests if an interface is empty, because an empty interface does
// not encode any information, we can't encode it in JSON so that the proxy
// knows it's a response, not a request.
func isNilValue(resp interface{}) bool {
```

This makes it unclear to me whether or not the conversion into a struct is
truly correct or not. I'm open to other ideas on how to resolve this issue.

### DIFF
--- a/langserver/definition.go
+++ b/langserver/definition.go
@@ -32,7 +32,7 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn JSONRPC2Conn, 
 		// Invalid nodes means we tried to click on something which is
 		// not an ident (eg comment/string/etc). Return no locations.
 		if _, ok := err.(*invalidNodeError); ok {
-			return nil, nil
+			return []symbolLocationInformation{}, nil
 		}
 		return nil, err
 	}
@@ -51,7 +51,7 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn JSONRPC2Conn, 
 			//
 			// TODO(sqs): find a way to actually emit builtin locations
 			// (pointing to builtin/builtin.go).
-			return nil, nil
+			return []symbolLocationInformation{}, nil
 		}
 	}
 	if len(nodes) == 0 {


slimsag:
@sqs I've tagged you in this PR because you wrote that code and it's very unclear from the docs why it's doing that. Please let me know what you think?

sqs:
Thanks for digging into this.

A response of `{"id":123,"result":{},"jsonrpc":"2.0"}` is a correct "no result found" response for some LSP methods, such as `textDocument/definition`. It was coincidental (well, somewhat of a convenience that no longer works) that `return nil, nil` mapped to that JSON-RPC response. I had `return nil, nil` to indicate "there is no definition found but there is also no unexpected error", as "no definition found" is not really considered an error at the LSP level. This was a fragile hack.

I think that if `handleDefinition` encounters "there is no definition found but there is also no unexpected error", it should return a sentinel error value like `ErrNoDefinition` that is translated into `{"id":123,"result":{},"jsonrpc":"2.0"}`, not a JSON-RPC error (`{"id":123,"error":{...whatever...},"jsonrpc":"2.0"}`), by the response handler. That means the handler should basically do `if err == ErrNoDefinition { return struct{}{}, nil }` in `textDocument/definition`.

> This makes it unclear to me whether or not the conversion into a struct is truly correct or not.

It is not. That is code I wrote and I see it is incorrect. An JSON-RPC 2.0 response's result can be any JSON type (http://www.jsonrpc.org/specification#response_object, note `result` has no constraints on type, unlike `params` in section 4.2). The `jsonrpc2.HandlerWithError` should return `{..., "result":null}` if the `result interface{}` return value is nil, and an LSP handler should never return `nil, nil` (because afaik `null` does not appear in any methods' specification as a valid value for the `result` field).

langserver/definition.go
```diff
@@ -51,7 +51,7 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn JSONRPC2Conn,
 			//
 			// TODO(sqs): find a way to actually emit builtin locations
 			// (pointing to builtin/builtin.go).
-			return nil, nil
+			return []symbolLocationInformation{}, nil
```
sqs: These changes are both correct.

sqs:
> I think that if handleDefinition encounters

Nevermind about this paragraph. I see that your change accomplishes the same result with less code.

felixfbecker:
`textDocument/xdefintition` must return an empty array if no result was found. Neither `null` or an empty object `{}` are valid return types. If there was an error, `result` must be omitted from the JSON, `null` is not valid either.

keegancsmith:
@felixfbecker FYI in go by default an empty slice (ie empty list) is nil.
```
var foo []string // foo is nil at this point
foo = append(foo, "hello world") // now foo is not nil

// instead you can do this, but this isn't idiomatic
foo := []string{}
```

PR #104, opened 2017-01-23
update to use new ServerCapabilities.TextDocumentSync field in LSP spec

sqs:
The server capabilities should be equivalent to what was sent previously, but they are sent using the new, richer TextDocumentSyncOptions.

### DIFF
--- a/pkg/lsp/service.go
+++ b/pkg/lsp/service.go
@@ -35,6 +35,10 @@ type InitializeError struct {
 	Retry bool `json:"retry"`
 }
 
+// TextDocumentSyncKind is a DEPRECATED way to describe how text
+// document syncing works. Use TextDocumentSyncOptions instead (or the
+// Options field of TextDocumentSyncOptionsOrKind if you need to
+// support JSON-(un)marshaling both).
 type TextDocumentSyncKind int
 
 const (
--- a/pkg/lsp/service.go
+++ b/pkg/lsp/service.go
@@ -1,5 +1,10 @@
 package lsp
 
+import (
+	"bytes"
+	"encoding/json"
+)
+
 type None struct{}
 
 type InitializeParams struct {
@@ -34,8 +39,8 @@ type TextDocumentSyncKind int
 
 const (
 	TDSKNone        TextDocumentSyncKind = 0
-	TDSKFull                             = 1
-	TDSKIncremental                      = 2
+	TDSKFull        TextDocumentSyncKind = 1
+	TDSKIncremental TextDocumentSyncKind = 2
 )
 
 type TextDocumentSyncOptions struct {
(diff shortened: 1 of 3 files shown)


pkg/lsp/service.go
```diff
@@ -38,8 +38,20 @@ const (
...
+	Save              *SaveOptions         `json:"save,omitempty"`
+}
+
+type SaveOptions struct {
+	IncludeText bool `json:"includeText"`
+}
+
 type ServerCapabilities struct {
-	TextDocumentSync                 int                              `json:"textDocumentSync,omitempty"`
+	TextDocumentSync                 TextDocumentSyncOptions          `json:"textDocumentSync,omitempty"`
```
keegancsmith: don't know this off the top of my head, but how does go handle unmarshalling something it expects to be an int, but it gets `TextDocumentSyncOptions` instead? Unless we do some sort of atomic rollout of this, that is gonna be an issue unless it just silently fails.

sqs:
Good point. Will need to add backcompat. Thanks for the catch.

pkg/lsp/service.go
sqs: @keegancsmith To address this, I made it so that either a JSON number (TextDocumentSyncKind) or an TextDocumentSyncOptions object can be JSON-unmarshaled into a TextDocumentSyncOptions Go struct. This means we can upgrade go-langserver first at any time.

sqs: Update: After my latest commit, we can update anything in any order, as long as we don't change anything to send  `TextDocumentSyncOptions` (need to keep it just sending the `TextDocumentSyncKind`). Once all are upgraded, then we can change them to send the `TextDocumentSyncOptions` in any order.

sqs:
@keegancsmith OK, ready for review again. No rush.

pkg/lsp/service.go
```diff
@@ -34,12 +39,67 @@ type TextDocumentSyncKind int
...
+	Change            TextDocumentSyncKind `json:"change"`
+	WillSave          bool                 `json:"willSave,omitempty"`
+	WillSaveWaitUntil bool                 `json:"willSaveWaitUntil,omitempty"`
+	Save              *SaveOptions         `json:"save,omitempty"`
+}
+
+// TextDocumentSyncOptions holds either a TextDocumentSyncKind or
+// TextDocumentSyncOptions. The LSP API allows either to be specified
+// in the (ServerCapabilities).TextDocumentSync field.
+type TextDocumentSyncOptionsOrKind struct {
```
keegancsmith: maybe document that the older type (number) is deprecated, but kept for backwards compatibility.

Man all these union types in the spec are really annoying in Go.

PR #105, opened 2017-01-23
Query.CaseSensitive support for symbols

mbana:
remove changes made to the langserver-go. (+5 squashed commits)
Squashed commits:
[074aeb5] remove un-needed files.
[2a396a4] `Query.CaseSensitive`:
remove reflection code and use the field name directly.
[ccab9ef] `Query`: fix typo in test.
[988dbe1] `Query`: fix typo should have been CaseSensitive.
[fae87b0] Query.CaseSensitive:
* add flag to indicate if Query is CaseSensitive.
* add tests for, the parsed Query string and the expected output. (+6 squashed commits)
Squashed commits:
[b865b75] set +x.
[8a9f71e] push to fork script.
[b7ad918] * configure tests as a typescript project.
*
[76c7255] * add install scripts to set remotes on the package in the $GOPATH.
* moving JS tests to the tests folder so you can load that in VSCode aswell.
[688cc32] update scripts to start websocket tests.
[a3a52e3] added websocket connect scripts using WebSocket class: run as $ ./scripts/websocket_test.sh

rename tool

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -28,10 +29,11 @@ import (
 // Query is a structured representation that is parsed from the user's
 // raw query string.
 type Query struct {
-	Kind      lsp.SymbolKind
-	Filter    FilterType
-	File, Dir string
-	Tokens    []string
+	Kind          lsp.SymbolKind
+	Filter        FilterType
+	File, Dir     string
+	Tokens        []string
+	CaseSensitive bool
 
 	Symbol lspext.SymbolDescriptor
 }
@@ -41,6 +43,9 @@ type Query struct {
 // structure back into a query string.
 func (q Query) String() string {
 	s := ""
+	if q.CaseSensitive {
+		s = queryJoin(s, fmt.Sprintf("casesensitive:%v", q.CaseSensitive))
+	}
 	switch q.Filter {
 	case FilterExported:
 		s = queryJoin(s, "is:exported")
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+npm-debug.log
+tests/js/out
```
keegancsmith: we don't need this gitignore file.

langserver/symbol.go
```diff
@@ -68,14 +72,40 @@ func queryJoin(s, e string) string {
...
+	// All queries are case insensitive, unless otherwise
+	qu, fields := parseCaseSensitive(q)
 
 	// Split the query into space-delimited fields.
-	for _, field := range strings.Fields(q) {
+	for _, field := range fields {
+		if !qu.CaseSensitive {
+			field = strings.ToLower(field)
+		}
+
```
keegancsmith: I think this code would be a lot clearer if you just added another check below which checks if a query is case sensitive. Then after this for loop you can adjust `qu` to be lower cased.

mbana: changed it around a bit and moved the conversion to a different func: `adjustCase`.
let me know if this clearer.

langserver/symbol.go
```diff
@@ -59,6 +62,7 @@ func (q Query) String() string {
 	for _, token := range q.Tokens {
 		s = queryJoin(s, token)
 	}
+	s = queryJoin(s, fmt.Sprintf("CaseSensitive:%v", q.CaseSensitive))
```
keegancsmith: this should go before the above for loop. It would also be nice that if q.CaseSensitive is the default value, that it isn't adding to the query string.

langserver/symbol.go
```diff
@@ -68,14 +72,50 @@ func queryJoin(s, e string) string {
...
+	return qu, fields
+}
+
 // ParseQuery parses a user's raw query string and returns a
 // structured representation of the query.
 func ParseQuery(q string) (qu Query) {
-	// All queries are case insensitive.
-	q = strings.ToLower(q)
+	// All queries are case insensitive, unless otherwise
+	qu, fields := adjustCase(q)
```
keegancsmith: I think you misunderstood my feedback earlier. You don't need this massive adjustCase function. You can just add another if condition in the below for loop to see if a field specifies case sensitivity. Then after that for loop act on that condition to adjust the `qu Query` struct.

langserver/symbol.go
mbana: i didn't want to change the default value.
how about `CaseInsensitive` instead? defaults to false, so not included, otherwise if true, add it.

langserver/symbol.go
mbana: yeh, ok that's cleaner. thanks.
so effectively, `qu = strings.toLower(qu.String())`?

keegancsmith: that wouldn't exactly work, but yes something like that.

langserver/symbol.go
```diff
@@ -71,11 +76,27 @@ func queryJoin(s, e string) string {
...
+	}
+	return qu
+}
 
+func doParseQuery(q string) (qu Query) {
 	// Split the query into space-delimited fields.
 	for _, field := range strings.Fields(q) {
+		// Check if the field is like `casesensitive:true`.
+		if strings.HasPrefix(field, "casesensitive:") {
+			if caseSensitive, err := strconv.ParseBool(strings.TrimPrefix(field, "casesensitive:")); err != nil {
```
mbana: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2232
wanted to inline there but couldn't - Non-name on left side.

keegancsmith:
Closing due to the age. If you are still keen to have this merged in I'll take another look.

PR #106, opened 2017-01-23
langserver: Extension to limit returned references

keegancsmith:


keegancsmith:
Seem like a reasonable extension?

felixfbecker:
Pagination support in general would make sense and I am pretty sure Microsoft will try to add something that works for all methods.

If you have `limit` you will likely want `offset` too.
Btw this reminds me that we forgot to prefix the workspace symbols parameter extension 😕

rothfels:
@felixfbecker there's still time to update that for `workspace/symbols`

keegancsmith:
@felixfbecker agreed. But in the interests of keeping it simple, I'd rather just have a limit with a large enough value that no user ever looks at that many. offsets makes life more difficult, so i'd prefer only adding that when we need it.

Also when microsoft adds pagination support to the protocol, we can just use that then. In the meantime we have very large responses being sent down the wire that the UI sometimes struggles to render on sourcegraph.com.

It may make more sense to do the Limit at another layer, but in general I don't like writing code in the proxy layer/etc which does special casing on the method.

felixfbecker:
Sounds good.

rothfels:
@keegancsmith please update the vendored code in the main repo

KattMingMing:
I'm passing `context: { includeDeclaration: false, xlimit: 100 },` for provideReferences

PR #107, opened 2017-01-23
langserver: return 'no results' not an error on builtins

slimsag:
This change causes textDocument/references on 'builtin' things (like `make` etc)
to return no results (empty array) instead of an error (`jsonrpc2: code 0 message: no package found for object builtin make`)

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -26,7 +26,7 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 		// Invalid nodes means we tried to click on something which is
 		// not an ident (eg comment/string/etc). Return no information.
 		if _, ok := err.(*invalidNodeError); ok {
-			return nil, nil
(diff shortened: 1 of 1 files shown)


langserver/references.go
```diff
@@ -54,6 +54,10 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	}
 
 	if obj.Pkg() == nil {
+		if _, builtin := obj.(*types.Builtin); builtin {
+			// TODO: support builtin with textDocument/references
```
nicot: Do you think we'll ever support builtins like this? Seems like it'd take a ton of computation.

keegancsmith: agreed. Rather just say lets not support it in this comment due to number of references.

slimsag:
Good points from both of you. Will update the docstring + merge

PR #109, opened 2017-01-23
refs: Find refs for xtest pkg definitions

keegancsmith:
Previously `TypeCheckFuncBodies` excluded xtest pkgs since `obj.Pkg().Path()`
would return the path without the `_test` suffix. I'm not sure why that is, but
normalising the pkg paths to not have _test allows find references to work.

### DIFF
--- a/langserver/langserver_test.go
+++ b/langserver/langserver_test.go
@@ -164,6 +164,19 @@ func TestServer(t *testing.T) {
 					"a_test.go:1:40": "var X int",
 					"a_test.go:1:46": "var A int",
 				},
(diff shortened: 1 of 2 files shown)


keegancsmith:
@sqs its probably a deficiency in upstream guru as well then. We should never mix up names, since we test for object equality via the position https://sourcegraph.com/github.com/sourcegraph/go-langserver@7f88185b9d334973d75de532fa6f68f992887ebe/-/blob/langserver/references.go#L218:1

So only objects defined in the same file and offset will be regarded as found, which is what we want.

PR #111, opened 2017-01-24
refs: Add timeout support

keegancsmith:
This may badly interact with using langserver in vscode, since large wait times
may be more acceptable. However, this solution is temporary until we add
cancellation support.

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -18,6 +18,7 @@ import (
 	"golang.org/x/tools/go/loader"
 	"golang.org/x/tools/refactor/importgraph"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sourcegraph/go-langserver/pkg/lsp"
 	"github.com/sourcegraph/jsonrpc2"
 )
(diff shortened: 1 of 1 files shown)


keegancsmith:
We could also make the timeout larger. We certainly want a timeout since in production we are limited by the timeout imposed by cloudflare (which is around 100s), but we can probably go higher. I'd like to see how this acts in practice though.

felixfbecker:
Is cancellation support a complicated one?

keegancsmith:
@felixfbecker not particularly, but this code would still have to exist anyways (and is easier). Once we do have cancellation support instead of using a `context.WithTimeout`, something higher up uses `context.WithCancel`.

langserver/references.go
```diff
@@ -8,19 +8,30 @@ import (
...
 	"github.com/sourcegraph/jsonrpc2"
 )
 
 func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lsp.ReferenceParams) ([]lsp.Location, error) {
+	// TODO: Add support for the cancelRequest LSP method instead of using
+	// hard-coded timeouts like this here.
+	//
+	// See: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#cancelRequest
+	// We just use the same timeout as workspace/xreferences
+	ctx, cancel := context.WithTimeout(ctx, workspaceReferencesTimeout)
```
slimsag: This shouldn't be `workspaceReferencesTimeout` but rather something else. Otherwise, rename that var to `referencesTimeout`

keegancsmith: I'll just make it something else, sorry for being lazy :P I do comment that `We just use the same timeout as workspace/xreferences`

PR #112, opened 2017-01-25
Return single result from definition handler

nicot:
The definition handler was returning list of definitions, when it could only return one.


keegancsmith:
> Union types will likely go away in the future from LSP.

Hallelujah. It was the most annoying part of the spec.

nicot:
Sounds good! I noticed that it could only return one definition, so I figured I'd remove the extra complexity.

PR #115, opened 2017-01-25
langserver: Add cancelRequest support

keegancsmith:
See spec at https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#cancelRequest

### DIFF
--- a/pkg/lsp/jsonrpc2.go
+++ b/pkg/lsp/jsonrpc2.go
@@ -1,0 +1,52 @@
+package lsp
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+// ID represents a JSON-RPC 2.0 request ID, which may be either a
+// string or number (or null, which is unsupported).
+type ID struct {
+	// At most one of Num or Str may be nonzero. If both are zero
+	// valued, then IsNum specifies which field's value is to be used
+	// as the ID.
+	Num uint64
+	Str string
+
+	// IsString controls whether the Num or Str field's value should be
+	// used as the ID, when both are zero valued. It must always be
+	// set to true if the request ID is a string.
+	IsString bool
+}
+
+func (id ID) String() string {
+	if id.IsString {
+		return strconv.Quote(id.Str)
+	}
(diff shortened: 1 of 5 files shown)


pkg/lsp/jsonrpc2.go
```diff
@@ -0,0 +1,52 @@
+package lsp
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+// ID represents a JSON-RPC 2.0 request ID, which may be either a
+// string or number (or null, which is unsupported).
+type ID struct {
```
keegancsmith: I add in this type rather than using `jsonrpc2.ID` to avoid this pkg having a dependency on `jsonrpc2`.

felixfbecker: What is the problem with the Go langserver having a dependency on jsonrpc2?

keegancsmith:
> What happens to a response when the request gets cancelled? Will the LS send the results it collected until the cancellation?

Yes we will still send a response. What the response is is up to the implementation of each method. They have to specifically listen for cancel signals to respond to them. Currently we only have two instances of those (workspace/xreferences and textDocument/references) and both of those return partial results.

felixfbecker:
sounds good. `workspace/symbol` should be cancellable too, VS Code does this automatically when you abort the search

keegancsmith:
Yeah will update the implementation for that. Should be easy to do, just hasn't been needed yet in sourcegraph.com production.

pkg/lsp/jsonrpc2.go
keegancsmith: This package in particular just contains LSP definitions, and has minimal dependencies. The language server isn't the only consumer of it, so it's common in go to do this sort of thing to avoid pulling in larger packages unnecessarily.

In fact it is a go proverb:
> A little copying is better than a little dependency.

https://go-proverbs.github.io/

felixfbecker: > A little copying is better than a little dependency.

Oh wow, the Go community is like the exact opposite of the NodeJS community

PR #118, opened 2017-01-26
WIP cancel typechecking upon context cancellation

slimsag:
This change causes us to:

1. Effectively abort type checking operations when the context deadline has occurred, thus freeing up valuable CPU and memory resources.
2. Properly cache FindPackage invocations made in textDocument/references requests.

### DIFF
--- a/langserver/build_context.go
+++ b/langserver/build_context.go
@@ -13,41 +13,10 @@ import (
 	"golang.org/x/net/context"
 )
 
-// withCancelContext creates a build.Context which wraps the input
-// *build.Context and aborts pending operations after ctx.Done
-func (h *LangHandler) withCancelContext(ctx context.Context, w *build.Context) *build.Context {
-	// We're mutating the build context that we intend to wrap, so copy it.
-	copy := *w
-	bctx := &copy
-
-	bctx.OpenFile = func(path string) (io.ReadCloser, error) {
-		if err := ctx.Err(); err != nil {
-			return nil, err
-		}
-		return w.OpenFile(path)
-	}
-	bctx.IsDir = func(path string) bool {
-		if err := ctx.Err(); err != nil {
-			return false
-		}
-		return w.IsDir(path)
-	}
-	bctx.HasSubdir = func(root, dir string) (rel string, ok bool) {
-		if err := ctx.Err(); err != nil {
-			return "", false
-		}
-		return w.HasSubdir(root, dir)
-	}
-	bctx.ReadDir = func(dir string) ([]os.FileInfo, error) {
-		if err := ctx.Err(); err != nil {
-			return nil, err
-		}
-		return w.ReadDir(dir)
-	}
-	return bctx
-}
-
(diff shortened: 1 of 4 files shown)


slimsag:
@keegancsmith Please carefully review https://github.com/sourcegraph/go-langserver/pull/118/commits/6b6dc25f20e586195786339eef95c2ef9ad0cabc -- I don't think there is anything wrong with it but I find it suspicious that it was not already the case. Did we disable that intentionally? e.g. to spare ourselves some memory usage for some reason?

langserver/build_context.go
```diff
@@ -13,6 +13,40 @@ import (
 	"golang.org/x/net/context"
 )
 
+// withCancelContext creates a build.Context which wraps the input
+// *build.Context and aborts pending operations after ctx.Done
+func (h *LangHandler) withCancelContext(ctx context.Context, w *build.Context) *build.Context {
+	// We're mutating the build context that we intend to wrap, so copy it.
+	copy := *w
+	bctx := &copy
```
keegancsmith: you can avoid the copy since we always use `withCancelContext`. I'd just inline in `BuildContext` and document there that if the context is cancelled we fail filesystem operations.

langserver/references.go
```diff
@@ -96,9 +97,21 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	} else {
 		users = h.importGraph.Search(defpkg)
 	}
+	findPackage := h.getFindPackageFunc()
```
keegancsmith: nice catch

langserver/references.go
```diff
@@ -96,9 +97,21 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
 	lconf := loader.Config{
 		Fset:  fset,
 		Build: bctx,
+		FindPackage: func(bctx *build.Context, importPath, fromDir string, mode build.ImportMode) (*build.Package, error) {
+			// When importing a package, ignore any
+			// MultipleGoErrors. This occurs, e.g., when you have a
+			// main.go with "// +build ignore" that imports the
+			// non-main package in the same dir.
+			bpkg, err := findPackage(ctx, bctx, importPath, fromDir, mode)
```
keegancsmith: Yeah there is a bug lurking here. Given this findPackage is a caching implementation, when we start failing operations due to the context being cancelled, we don't want to cache those results.

This is a little tricky to fix. Two options:
* Document that we can fail findpackage FS operations and update our private implementation not to cache if the context is cancelled after the operation is done
* Not do this approach of failing bctx.

In general, BuildContext is used everwhere, and we cache failure in most places. So this approach actually introduces a bunch of bugs around caching.

slimsag: (I'd like to chat about this Mon to get your perspective, I think I'm missing exact details here / I have a few questions)

keegancsmith:
Shall we abandon this PR in the meantime?

PR #119, opened 2017-01-26
support *_test.go files inside of workspace/xreferences

slimsag:
See commit descriptions for more details

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -67,6 +67,7 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
 			}
 		}
 		unvendoredPackages[bpkg.ImportPath] = struct{}{}
+		unvendoredPackages[bpkg.ImportPath+"_test"] = struct{}{}
 		pkgs = append(pkgs, pkg)
 	}
 	if len(pkgs) == 0 {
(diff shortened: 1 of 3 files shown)


langserver/workspace_refs.go
```diff
@@ -67,6 +67,7 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
 			}
 		}
 		unvendoredPackages[bpkg.ImportPath] = struct{}{}
+		unvendoredPackages[bpkg.ImportPath+"_test"] = struct{}{}
```
keegancsmith: I was expecting the main change to be just the `ImportWithTests` like we do in textDocument/refereces. Do you also do this to avoid tricks like `strings.TrimSuffix(pkg, "_test")` that we do in the other reference code?

slimsag: Right, so if you find the actual usage of `unvendoredPackages` below 86-90 you'll see that this is a map of packages that we're interested in scanning for external references (b/c `AfterTypeCheck` is invoked for transitive dependencies too). A better name _might_ be "interestingUnvendoredPackages" but that's a bit lengthy already, hence leaving it as-is.

slimsag: This is roughly the same as doing `strings.TrimSuffix(pkg, "_test")` inside the other reference code.

PR #120, opened 2017-01-27
langserver: clear *loader.PackageInfo fields after use (lower memory consumption)

slimsag:


langserver/workspace_refs.go
```diff
@@ -101,6 +102,7 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
 			if err != nil {
 				log.Printf("workspaceRefsFromPkg: %v: %v", pkg, err)
 			}
+			clearInfoFields(pkg) // save memory
```
keegancsmith: you can do this in the defer above, so we still free memory even if we panic.

PR #121, opened 2017-01-27
symbol: Only inspect pkg in symbol.id

keegancsmith:
If symbol.id is present, we know we only have to inspect that pkg. This should
greatly speed up symbol queries using this extension on large repos.


felixfbecker:
That is exactly the kind of optimization that `id` is supposed to do 👍

Issue #85, opened 2017-01-13
consider a more structured code layout


slimsag:
From my very harsh/critical perspective, `go-langserver` is rapidly becoming a giant glob of code living in a single directory. This is bad because our code documentation can rapidly become out of date or interconnected pieces can be altered affecting other things unintentionally.

This is less of a "I have a concrete plan" and more of a "let's chat about how we can improve this" issue. Ideally we have a top-down polytree structure without any cyclic dependencies etc.

One good example is how `langserver/internal/refs` is totally separate from the main package. Granted, more of the external references code in `workspace_refs.go` can and should be moved away from the main `langserver` package -- but it's a good reference idea here.

Take this as a plea from me to find and place code that is sufficiently independent and factor it out into smaller separate `internal` subpackages @keegancsmith :) (I'll be trying to do this actively, too, obviously)

keegancsmith:
I agree, its organic growth has led to code that could be better structured, so we should do some improvements. I'm not sure if subpackages (excluding for larger bits of code such as refs or symbols) would actually make life easier though. I think the bigger win would be better thought put into how all the shared state is accessed/separated out. Right now we have stuff split out over files in a haphazard way, but it is all essentially doing things to `LangHandler`. So that seems like a good win rather than just moving code around.

BTW I don't see our code layout as a bad thing. I'm glad we got to this point of confusing code! Because now I think it is more obvious how we can improve the codebase, than if we designed it upfront when all we supported was hover, ref and j2d. But yes, prioritizing some refactoring/cleanup work is super important, especially if we want more outside contributors to not just run away when they open up this codebase.

keegancsmith:
I'm going to try ship a few small changes this week which should start shaping us to a codebase that makes more sense. The mini-goal I set this week is making https://godoc.org/github.com/sourcegraph/go-langserver/langserver look reasonable. The first PR is here https://github.com/sourcegraph/go-langserver/pull/87

PR #124, opened 2017-01-30
RFC WIP Move init/shutdown to another server

keegancsmith:
This a WIP PR to get feedback on this approach. One of the most annoying things about go-langserver right now is all the handler fields have to be protected by a mutex, since we do not know if they have been init yet. It is also written that way since we have to handle the server being shutdown and reused for another workspace.

Instead this PR pulls Init and Shutdown into another handler. It has a function Init which is just like Handle, except it will only ever be called for initialize and it is expected to return a jsonrpc2.Handler which can then handle usual requests. The reason this is an improvement is the handler Init returns doesn't ever have to worry about `Reset`/etc, so we can pretty much get rid of the large amount of locking we do whenever we access something.

Related to https://github.com/sourcegraph/go-langserver/issues/85

### DIFF
--- a/pkg/lspserver/lspserver.go
+++ b/pkg/lspserver/lspserver.go
@@ -1,0 +1,67 @@
+// package lspserver implements a general LSP server.
+package lspserver
+
+import (
+	"context"
+	"sync"
+
+	"github.com/sourcegraph/jsonrpc2"
+)
+
+type Handler struct {
+	Init func(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrpc2.Request) jsonrpc2.Handler
+
+	mu sync.Mutex
+	h  jsonrpc2.Handler
+}
+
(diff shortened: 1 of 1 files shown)


keegancsmith:
I'll try integrate this with the actual go-langserver at another time. But I think the above comment + code should help communicate the important idea here. We can also move things like support for cancellation/tracing into this server, and make the go-langserver just handle workspace/* and textDocument/* methods.

This also needs to be exported, since we will want to use it in our build_server. As part of that change it will probably also lead to improving the situation around having 3 handler structs.

slimsag:
Overall I like it! I think perhaps long term we will find that we don't need a separate layer like done here, but I don't have a concrete idea on how that would look at the moment. It's clear to me this would be a big improvement though.

Assuming some more docs are added to the pkg (`lspserver` seems like a lame name, too) then SGTM!

keegancsmith:
Abandon ship

PR #125, opened 2017-01-30
refs: Stream back partial results

keegancsmith:
This is an experiment for streaming back reference results, before the reference
operation is finished. It uses an unspecced notification called
`$/partialResult` which is a list of JSONPatch operations to apply. We stream
back results every second if we have any unsent results.

Currently the order results are sent back in is very different to what the final
result will be. It will also ignore the limit specifier. We can adjust this in
future as we see how the prototype behaves.

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -9,6 +9,7 @@ import (
 	"go/token"
 	"go/types"
 	"log"
+	"os"
 	"path"
 	"sort"
 	"strings"
@@ -20,12 +21,15 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sourcegraph/go-langserver/pkg/lsp"
+	"github.com/sourcegraph/go-langserver/pkg/lspext"
 	"github.com/sourcegraph/jsonrpc2"
 )
 
 // documentReferencesTimeout is the timeout used for textDocument/references
 // calls.
 const documentReferencesTimeout = 15 * time.Second
+
+var streamExperiment = len(os.Getenv("STREAM_EXPERIMENT")) > 0
 
 func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSONRPC2Conn, req *jsonrpc2.Request, params lsp.ReferenceParams) ([]lsp.Location, error) {
 	// TODO: Add support for the cancelRequest LSP method instead of using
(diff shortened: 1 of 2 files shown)


pkg/lspext/partial.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "github.com/sourcegraph/go-langserver/pkg/lsp"
+
+// PartialResultParams is the input for "$/partialResult", a notification.
+type PartialResultParams struct {
+	ID      lsp.ID      `json:"id"`
+	Patches []JSONPatch `json:"patches"`
+}
+
+type JSONPatch struct {
```
felixfbecker: This is an `AddOperation`, other operations have a different schema

langserver/references.go
```diff
@@ -181,10 +185,51 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+		patches := make([]lspext.JSONPatch, len(partial)-streamPos, 0)
+		for ; streamPos < len(partial); streamPos++ {
+			patches = append(patches, lspext.JSONPatch{
+				OP:    "add",
+				Path:  "/-",
+				Value: goRangeToLSPLocation(fset, partial[streamPos].Pos(), partial[streamPos].End()),
+			})
+		}
+		if initial {
+			patches[0].Path = ""
```
felixfbecker: This only sets the path, what about the value?
The first operation must be `{"op": "add", "path": "", "value": []}`, to initialize the result to an empty array on the client. Otherwise the client will not be able to apply further pushes the array, because the result will not be an array, but `undefined`.

pkg/lspext/partial.go
felixfbecker: We might wanna move these out into a different file for JSON Patch.
have you looked at https://github.com/evanphx/json-patch and https://github.com/mattbaird/jsonpatch?

keegancsmith: Our current use case is super simple, we don't need patching algorithms in the go-langserver.

langserver/references.go
```diff
@@ -181,10 +185,64 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+		mu.Unlock()
+		if len(partial) == streamPos {
+			continue
+		}
+
+		type AddOp struct {
+			OP    string
+			Path  string
+			Value lsp.Location
+		}
```
keegancsmith: I need to update the JSON fields for this

pkg/lspext/partial.go
```diff
@@ -0,0 +1,16 @@
...
+type PartialResultParams struct {
+	// ID is the jsonrpc2 ID of the request we are returning partial
+	// results for.
+	ID lsp.ID `json:"id"`
+
+	// Patch is a JSON patch as specified at http://jsonpatch.com/
+	//
+	// It is an interface{}, since our only requirement is that it JSON
+	// marshals to a valid list of JSON Patch operations.
+	Patch interface{} `json:"patch"`
```
felixfbecker: Is there a downside to enforcing this to an array?

keegancsmith: Yes, go doesn't have proper generics. `[]interface{}` is not the same as `[]AddOp` / can't be autocast.

keegancsmith:
> What are you returning as the final response?

I haven't changed the final response at all. FYI In general the final result will have a very different order to how we send out the patches.

felixfbecker:
> I haven't changed the final response at all.

That is good for BC.

> In general the final result will have a very different order to how we send out the patches.

Why? Either the result is considered unsorted, in which case the LS should return the result that has been "built" through JSON patches, or the result is sorted, in which case the LS should add the items at the right position in the result array.

keegancsmith:
* I don't think applying all partialResult patches should result in the final thing that is returned. Otherwise one just need to send a final `partialResult` notification before sending the final result, which seems uneeded.
* The result is sorted, but we can only do that once we have all of them. Sending patches down which insert at  the "correct" positions is possible, but I think unnecessary and undesirable. Lets see what happens when we have this working. (Undesirable because I feel like ref results jumping around/disappearing would be a bad UI)

felixfbecker:
> Otherwise one just need to send a final partialResult notification before sending the final result, which seems uneeded.

In the future if the client opts in that it supports partialResult notifications the final response could just `result: null` to save the bandwidth. It would just mark the result as completed.

 > Lets see what happens when we have this working. (Undesirable because I feel like ref results jumping around/disappearing would be a bad UI)

But is exactly what would happen if the final result has a totally different order. The client would use the final result and suddenly all list items would jump around. Or the client would ignore the final result, in which case the list would be unsorted.

Btw, does LSP even define that the result must be sorted? I don't do any sorting in PHP on the server for references

keegancsmith:
We can remove sorting. I originally added it back in the day because vscode showed the results in a non-optimal order. It may have improved the code since then. FYI the order tries to make it so that results that are "closer" to the position you clicked appear closer to the front.

felixfbecker:
We need to make sure that the client applies sorting then, since we now render the references ourselves in the sidebar. But the final result should be consistent with the result of the patches.

keegancsmith:
Yup, we will ensure that. I'm gonna merge this now, we can followup for other changes in a new PR.

PR #126, opened 2017-01-30
references: Send a final partialResult

keegancsmith:
This will allow clients to ignore the actual result, and just rely on
partialResult.

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -197,7 +197,7 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 		lconf.Load() // ignore error
 	}()
 
-	streamPos := 0
+	streamUpdate := func() {}
 	streamTick := make(<-chan time.Time, 1)
 	if streamExperiment {
 		initial := json.RawMessage(`[{"op":"add","path":"","value":[]}]`)
@@ -212,6 +212,32 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 		t := time.NewTicker(time.Second)
 		defer t.Stop()
 		streamTick = t.C
+		streamPos := 0
+		streamUpdate = func() {
+			mu.Lock()
+			partial := refs
+			mu.Unlock()
+			if len(partial) == streamPos {
+				return
+			}
+
+			patch := make([]ReferenceAddOp, 0, len(partial)-streamPos)
+			for ; streamPos < len(partial); streamPos++ {
+				patch = append(patch, ReferenceAddOp{
+					OP:    "add",
+					Path:  "/-",
+					Value: goRangeToLSPLocation(fset, partial[streamPos].Pos(), partial[streamPos].End()),
+				})
+			}
+			conn.Notify(ctx, "$/partialResult", &lspext.PartialResultParams{
+				ID: lsp.ID{
+					Num:      req.ID.Num,
+					Str:      req.ID.Str,
+					IsString: req.ID.IsString,
+				},
+				Patch: patch,
+			})
+		}
 	}
 Loop:
 	for {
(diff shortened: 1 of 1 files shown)


keegancsmith:
Suggest reviewing with `?w=1`

felixfbecker:
What exactly does this mean?

keegancsmith:
We didn't send every location back before this PR. Any location found between the last time we sent the partialResult and the typechecker finishing were not sent.

Atleast for our prototype implementation, we are going to ignore the result of `testDocument/reference`, and instead only rely on `$/partialResult`.

felixfbecker:
> Any location found between the last time we sent the partialResult and the typechecker finishing were not sent.

Don't fully understand that sentence.

> Atleast for our prototype implementation, we are going to ignore the result of testDocument/reference, and instead only rely on $/partialResult.

That is totally fine and even desirable for forward compatibility.

slimsag:
> Don't fully understand that sentence.

It means that e.g. we would send `$/partialResult` periodically while the request was in-flight, but once the request was considered "done" by the server, we didn't send the final `$/partialResult` and thus would occasionally leave off the very last/final partial result. This fixes it.

langserver/references.go
```diff
@@ -212,6 +212,32 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 		t := time.NewTicker(time.Second)
 		defer t.Stop()
 		streamTick = t.C
+		streamPos := 0
+		streamUpdate = func() {
+			mu.Lock()
+			partial := refs
+			mu.Unlock()
+			if len(partial) == streamPos {
```
slimsag: What happens if `partial` (which is really `refs` cloned) changes behind the scenes? Is returning here important?

langserver/references.go
```diff
@@ -212,6 +212,32 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+				return
+			}
+
+			patch := make([]ReferenceAddOp, 0, len(partial)-streamPos)
+			for ; streamPos < len(partial); streamPos++ {
+				patch = append(patch, ReferenceAddOp{
+					OP:    "add",
+					Path:  "/-",
```
slimsag: SG only shows me the usage here, so, what does `Path` do out of curiosity?

langserver/references.go
```diff
@@ -212,6 +212,32 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+					Value: goRangeToLSPLocation(fset, partial[streamPos].Pos(), partial[streamPos].End()),
+				})
+			}
+			conn.Notify(ctx, "$/partialResult", &lspext.PartialResultParams{
+				ID: lsp.ID{
```
slimsag: Since this isn't a pointer type can't you just do `ID: req.ID`?

langserver/references.go
felixfbecker: It's where the item should be added in the object.
See http://jsonpatch.com/ and https://tools.ietf.org/html/rfc6901 for the spec

slimsag:
Some minor comments / questions -- feel free to address as you see fit

langserver/references.go
felixfbecker: In this case, it means at the end of the array

langserver/references.go
keegancsmith: this line means everything in refs so far has already been sent. If refs changes after mu.Unlock but before this line, partial won't change. We need to snapshot the slice (note its just the slice "header", not the actual content of the slice so is small) since reading `refs` outside of the lock is a concurrent read/write (ie undefined behaviour). Holding the lock while we build `patch` is much longer than we need to hold the lock.

langserver/references.go
keegancsmith: `lsp.ID` is different to `req.ID` which is a `jsonrpc2.ID`. They are actually the same, but I avoided using `jsonrpc2.ID` in `lsp.ID` to avoid adding `jsonrpc2.ID` as a dep. FYI in the version of Go coming out soon I'd just be able to do `ID: lsp.ID(req.ID)`.

langserver/references.go
slimsag: Maybe add a comment `// Everything in refs has already been sent, so return.` ? that would've answered my Q here :)

langserver/references.go
slimsag: Gotcha, thank you!

langserver/references.go
slimsag: Nice, I missed that :)

PR #127, opened 2017-01-30
langserver: do not use goroutines during AfterTypeCheck

slimsag:
The use of them prevents us from invoking `clearInfoFields(pkg)` and saving
memory (as seen in https://github.com/sourcegraph/go-langserver/pull/122)
and also can cause concurrent map accesses of `pkg.Info*`

### DIFF
--- a/langserver/workspace_refs.go
+++ b/langserver/workspace_refs.go
@@ -79,10 +79,7 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
 	// Collect dependency references in the AfterTypeCheck phase. This enables
 	// us to begin looking at packages as they are typechecked, instead of
 	// waiting for all packages to be typechecked (which is IO bound).
-	var (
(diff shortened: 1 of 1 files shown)


langserver/workspace_refs.go
```diff
@@ -79,31 +79,23 @@ func (h *LangHandler) handleWorkspaceReferences(ctx context.Context, conn JSONRP
...
 	// waiting for all packages to be typechecked (which is IO bound).
-	var (
-		results = refResultSorter{results: make([]referenceInformation, 0)}
-		wg      sync.WaitGroup
-	)
+	var results = refResultSorter{results: make([]referenceInformation, 0)}
 	afterTypeCheck := func(pkg *loader.PackageInfo, files []*ast.File) {
 		_, interested := unvendoredPackages[pkg.Pkg.Path()]
 		if !interested {
-			clearInfoFields(pkg) // save memory
```
keegancsmith: you still want to do this

slimsag: Whoops, you're right.

PR #129, opened 2017-01-31
reference: Initial stream op is a replace

keegancsmith:
This is a more robust way to start the stream. For example, if there is an
automatic retry at some layer (true for sourcegraph.com), this will allow a
retried stream to start from scratch again.

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -200,7 +200,7 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	streamUpdate := func() {}
 	streamTick := make(<-chan time.Time, 1)
 	if streamExperiment {
-		initial := json.RawMessage(`[{"op":"add","path":"","value":[]}]`)
+		initial := json.RawMessage(`[{"op":"replace","path":"","value":[]}]`)
 		conn.Notify(ctx, "$/partialResult", &lspext.PartialResultParams{
 			ID: lsp.ID{
 				Num:      req.ID.Num,


keegancsmith:
cc @nicksnyder this is a possibility you may encounter in the frontend. Receiving a stream of references, and then it getting reset and starting streaming again. This happens due to automatic retries in our lsp-proxy.

felixfbecker:
The JSON Patch spec defines that a replace is equivalent to a remove followed by an add, so this is fine and a good idea.

But how can it happen that the same request partial result stream gets "resetted"? If there is an error, it should return an error response, and the client can choose to retry as a new request with a new request ID. This could get very messy otherwise.

keegancsmith:
@felixfbecker yes we should consider retries only at the UI layer. However there are some advantages to the current approach:
* The lsp-proxy knows if a failure was a transient network error (the only reason we retry)
* Retries at the lsp-proxy layer are faster (data center networking/feedback loop)
* All operations we proxy are immutable (ie we do not proxy requests which could mutate the state of the langserver)
* Retries currently are effective and rare (a handful of times a day)
* A request hitting our http frontend represents what we show the user, which makes for much more effective measurements of our reliability.

Streaming is the first example of something which leaks that we do transparent retries. Given the pros of our current retry approach I'm keen to keep it until it proves to be bad. In either case, the frontend will have to handle streams being interrupted. In this case it will be an empty replace op, in the other case a failed response + retry.

felixfbecker:
Okay, we just have to consider:
- We just have to make sure all LS behave this way
- It might cause a weird UX if the result list suddenly resets without the client knowing it came because of an error condition/retry. It would make a better UX to show an error indicator and/or keep the partial result list with an explicit retry button etc. The proxy could return a 502 Bad Gateway or 504 Gateway Timeout.

I would assume a network error between the proxy and LS is pretty rare, right?

keegancsmith:
> I would assume a network error between the proxy and LS is pretty rare, right?

that is exactly what the retry is on. Network errors also happen due to race conditions on deployments/OOMs/etc. It happens a few times a week, but in those cases helps us recover quite well. You can see a graph of retries vs retries that failed here https://grafana.sgdev.org/dashboard/db/xlang-go?panelId=8&fullscreen&from=now-3h&to=now

nicksnyder:
If a retry succeeds, we probably don't expect the results to be different. In this case, clearing the results on reset seems less than ideal from a user perspective.

What if we kept the old results and just continued appending to them (using replace instead of add)?

keegancsmith:
the langserver failing is usually a sign it died and we are using another one.
* the other server has no idea how far the stream is
* the order we find and stream results is not consistent
* the idea of the replace op is to signal to the UI that a retry is happening. It should probably be smart about merging in the results given they will be the same, just in a different order. The same logic would have to exist if retries moved to the UI (if we want to avoid bad UX)

felixfbecker:
Agree with the first two points, disagree with the last. The client should not assume a retry in the case of a replace, it should just apply the patches as they come in. If we want error handling, we should switch to reporting an error and letting the client retry.

nicksnyder:
@keegancsmith 

> the other server has no idea how far the stream is
> the order we find and stream results is not consistent

I don't quite understand why these constraints would prevent using replace semantics everywhere. Lets walk thought a hypothetical example:

1. Client sends ref request for a symbol Foo.
2. There are currently 3 references to Foo, each in a different repo: Alpha, Bravo, Charlie.
3. The langserver starts streaming references through the proxy and to the browser (Alpha.Foo, Bravo.Foo) but gets disconnected before sending Charlie. Note that even on this request it should be fine to use "replace" instead of "add" for all results.
4. The proxy retries to another langserver. That langserver has no knowledge of the state from the previous langserver.
5. The new langserver starts streaming back results in a different order (using replace instead of add for all results)
    a. Bravo.Foo (replaces existing reference that the browser has)
    b. Charlie.Foo (new reference)
    c. Alpha.Foo (replaces existing reference that the browser has)

I think the proxy can be completely stateless w.r.t. the result state, just forwarding results that it streams.

Does that make sense? Am I missing something about how the proxy and/or langserver work?

@felixfbecker I agree. If my suggestion works, then the UI/browser doesn't even need to know that a retry is happening. It just keeps happily applying diffs. Some of the diffs may be redundant (e.g. Bravo.Foo, Alpha.Foo), but that is better user experience than deleting them and waiting for the new langserver to return them.

felixfbecker:
@nicksnyder Are you suggesting the LS should never add to the list, but always _replace_ the whole list with an updated list that has one more member? When you say "for all results", do you mean "for all patches"? That is unnecessary overhead and goes against the whole idea streaming is trying to solve. For 10,000 references you would send (sum n=1 to 10,000) = 50,005,000 references in total over the wire.

nicksnyder:
No, that is not what I want. I want individual references to stream one by one all the way from the langserver to the browser. Is that not possible? Does the proxy have to poll the langserver?

nicksnyder:
I don't have an understanding of the details of how t currently works, but I want to replace elements in the list of references, not replace the list of references itself.

felixfbecker:
It works like this
- ➡️  request `workspace/xreferences`, ID 1
- ⬅️  `$/partialResult` for ID 1, replace root with `[]`
- ⬅️  `$/partialResult` for ID 1, add to `/-` the item Alpha
- ⬅️  `$/partialResult` for ID 1, add to `/-` the item Bravo
- ⬅️  `$/partialResult` for ID 1, add to `/-` the item Charly
- ⬅️   response for ID 1 is Alpha, Bravo, Charly
  (in the future this does not have to include the final result)

nicksnyder:
@felixfbecker and I chatted on the phone.

We concluded that although it may be technically possible to add logic to the proxy to transparently handle langserver retries in a way that prevents the browser from reseting its state, it isn't worth doing because langserver retries are rare and the complexity would be non-trivial.

Thanks for having the discussion, it helped me understand the current state of the system.

PR #138, opened 2017-02-06
references: Make it possible to use incremental import graphs

keegancsmith:
This change is mostly refactoring and cleanup. It is intended to make it possible to use inaccurate import graphs, as well as making it possible to do multiple passes with more accurate import graphs. I have made some changes to `findReferences` which makes it responsibility clearer as well.

The follow up PR for this will actually implement the major perf improvement of using a cached import graph. I left it out of this PR in the interest of keeping this smaller.

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -205,9 +205,15 @@ func (h *LangHandler) reverseImportGraph() <-chan importgraph.Graph {
 			findPackage := func(bctx *build.Context, importPath, fromDir string, mode build.ImportMode) (*build.Package, error) {
 				return findPackageWithCtx(ctx, bctx, importPath, fromDir, mode)
 			}
+			h.mu.Lock()
 			h.importGraph = tools.BuildReverseImportGraph(bctx, findPackage, h.FilePath(h.init.RootPath))
+			h.mu.Unlock()
 		})
-		c <- h.importGraph
+		h.mu.Lock()
+		// TODO(keegancsmith) h.importGraph may have been reset after once
+		importGraph := h.importGraph
+		h.mu.Unlock()
+		c <- importGraph
 
 		close(c)
 	}()
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -42,6 +42,8 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	ctx, cancel := context.WithTimeout(ctx, documentReferencesTimeout)
 	defer cancel()
 
+	reverseImportGraphC := h.reverseImportGraph()
+
 	fset, node, _, _, pkg, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
 	if err != nil {
 		// Invalid nodes means we tried to click on something which is
@@ -50,14 +52,6 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 			return []lsp.Location{}, nil
 		}
 		return nil, err
-	}
-
-	// TODO use successive import graphs. For now we just use the last and
-	// most accurate import graph.
-	reverseImportGraphC := h.reverseImportGraph()
-	var reverseImportGraph importgraph.Graph
-	for g := range reverseImportGraphC {
-		reverseImportGraph = g
 	}
 
 	// NOTICE: Code adapted from golang.org/x/tools/cmd/guru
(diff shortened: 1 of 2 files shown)


langserver/references.go
```diff
@@ -263,15 +343,6 @@ func findReferences(ctx context.Context, lconf loader.Config, collectPkg func(st
...
 		defer wg.Done()
 
-		if ctx.Err() != nil {
-			// work has been cancelled. We need to check this
-			// after wg.Add since below we do a wg.Wait after
-			// finding out the context is cancelled. Otherway
-			// around could lead to us sending down a closed refs
-			// channel.
-			return
-		}
```
keegancsmith: this we removed since `TypeCheckFuncBodies` checks this as well

langserver/references.go
```diff
@@ -166,6 +186,44 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+			findPackageWithCtx := h.getFindPackageFunc()
+			findPackage := func(bctx *build.Context, importPath, fromDir string, mode build.ImportMode) (*build.Package, error) {
+				return findPackageWithCtx(ctx, bctx, importPath, fromDir, mode)
+			}
+			h.mu.Lock()
+			h.importGraph = tools.BuildReverseImportGraph(bctx, findPackage, h.FilePath(h.init.RootPath))
+			h.mu.Unlock()
+		})
+		h.mu.Lock()
+		// TODO(keegancsmith) h.importGraph may have been reset after once
```
keegancsmith: Part of my tasks this week is to make go-langserver much better at handling mutations. As part of that work I will be fixing up TODOs like this.

langserver/references.go
```diff
@@ -166,6 +166,38 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+// graph sent is accurate. The reader does not have to read all the values.
+func (h *LangHandler) reverseImportGraph() <-chan importgraph.Graph {
+	// Ensure our buffer is big enough to prevent deadlock
+	c := make(chan importgraph.Graph, 1)
+
+	// Note: We use a background context since this operation should not
+	// be linked to an individual request.
+	ctx := context.Background()
+
+	go func() {
```
slimsag: Any situation in which this could panic?

langserver/references.go
```diff
@@ -42,6 +42,8 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 	ctx, cancel := context.WithTimeout(ctx, documentReferencesTimeout)
 	defer cancel()
 
+	reverseImportGraphC := h.reverseImportGraph()
```
slimsag: As it stands it might not be obvious why this is here (at the top of the function) instead of at the bottom where it is used. Add a comment:

```
// Begin computing the reverse import graph immediately, as this occurs in the background
// and is IO-bound.
```

langserver/references.go
```diff
@@ -85,6 +85,23 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
...
+
+		// refs is a stream of raw references found findReferences.
+		refs = make(chan *ast.Ident)
+
+		// findRefErr is non-nil if findReferences fails.
+		findRefErr error
+	)
+
+	go func() {
+		locsC <- refStreamAndCollect(ctx, conn, req, fset, refs)
```
slimsag: Should probably have a panic handler here.

langserver/references.go
```diff
@@ -142,7 +157,10 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn JSO
 		log.Printf("info: timeout during references for %s, found %d refs", defpkg, len(locs))
 	}
 
-	if findRefErr != nil {
+	// If we find references then we can ignore findRefErr. It should only
+	// be non-nil due to timeouts or our last findReferences doesn't find
+	// the def.
+	if len(locs) == 0 && findRefErr != nil {
```
slimsag: Does this mean an LSP error occurs if we find zero references? That doesn't sound like an error to me.

langserver/references.go
keegancsmith: this really should never panic, we control all the code in it, including libraries it calls. I'd rather it take down the server so we notice it.

Other places we have panic handlers in goroutines is due to using `go/types`/etc libraries on all the code out in the wild.

langserver/references.go
keegancsmith: `findRefErr` can be non-nil (ie it could be `ctx.Err()` or an error typechecking `defpkg`). However, if we found references and it is non-nil that means:
* context timeout, we want partial results.
* we could do findReferences more than once. The last check could fail, but an earlier one succeed. We should still return results in such a case.

keegancsmith:
@slimsag most of the changes are actually just moving code around. Most commits are minimal when viewing without whitespace changes. The only functional change is we now compute the reverse import graph sooner and in the background.

Our test suite is actually not that big in the case of references. However, it covers the essentials. Also there is very minimal functional changes in this as I mentioned. I heavily tested `references: Refactor to only use lconf.ImportPkgs`, which is the only real change (even though it should work exactly as it did before).

However, I will monitor + manual test when I roll out.

PR #139, opened 2017-02-06
references: Use xcache to store older importgraphs

keegancsmith:
This should greatly speed up initial reference requests in the common case. In
sourcegraph.com production this would make us use a possibly inaccurate
importgraph to find references (computed from another commit). We still follow
up with the correct import graph, but only typecheck the package we had not
already checked. With this change and streaming, references should hopefully
almost instantaneously start streaming back, even on large repos.

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -229,9 +229,7 @@ func (h *LangHandler) reverseImportGraph(conn jsonrpc2.JSONRPC2) <-chan importgr
 			h.mu.Unlock()
 
 			// Update cache in background
-			go func() {
-				h.cacheSet(ctx, conn, cacheKey, g)
-			}()
+			go h.cacheSet(ctx, conn, cacheKey, g)
 		})
 		h.mu.Lock()
 		// TODO(keegancsmith) h.importGraph may have been reset after once
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -44,7 +44,7 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 
 	// Begin computing the reverse import graph immediately, as this
 	// occurs in the background and is IO-bound.
-	reverseImportGraphC := h.reverseImportGraph()
+	reverseImportGraphC := h.reverseImportGraph(conn)
 
 	fset, node, _, _, pkg, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


langserver/references.go
```diff
@@ -211,6 +227,11 @@ func (h *LangHandler) reverseImportGraph() <-chan importgraph.Graph {
 			h.mu.Lock()
 			h.importGraph = g
 			h.mu.Unlock()
+
+			// Update cache in background
+			go func() {
+				h.cacheSet(ctx, conn, cacheKey, g)
+			}()
```
slimsag: This can just be `go h.cacheSet(tx, conn, cacheKey, g)` no?

langserver/xcache.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	var r json.RawMessage
+	err := conn.Call(ctx, "xcache/get", lspext.CacheGetParams{Key: key}, &r)
+	if err != nil {
+		return false
+	}
+	b := []byte(r)
+	if bytes.Equal(b, []byte("null")) {
+		return false
```
slimsag: This could be written more simply as `if string(r) == "null" {` I think.

langserver/references.go
```diff
@@ -190,16 +190,32 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
...
 
 	// Note: We use a background context since this operation should not
 	// be linked to an individual request.
 	ctx := context.Background()
 
 	go func() {
-		// TODO fetch from cache an inaccurate import graph
+		// This should always be related to the go import path for
+		// this repo. For sourcegraph.com this means we share the
+		// import graph across commits. We want this behaviour since
```
felixfbecker: I think you mean _import path_

keegancsmith: nope, I mean import graph

felixfbecker: k, looked like a typo

langserver/xcache.go
keegancsmith: That would cause an allocation, since []byte (which json.RawMessage is a type-alias to) is mutable and string is immutable.

slimsag: Yep, I'm aware, but it would be quite small (4 bytes) and imho worth the readability. Anywho, it's OK either way 😄

felixfbecker:
XCache FTW

langserver/xcache.go
keegancsmith: r is not necessarily 4 bytes, it may be very large.

slimsag: Oh, right, wow, I missed that :/ thank you

PR #140, opened 2017-02-08
references: Find refs to symbols defined in xtest

keegancsmith:
We just need to treat `foo_test` as if it is `foo`. We do that in a few places,
but we now do it everywhere. I was concerned about possibly reporting duplicate
results, but our unit tests indicate that is not a concern.

### DIFF
--- a/langserver/langserver_test.go
+++ b/langserver/langserver_test.go
@@ -158,6 +158,10 @@ func TestServer(t *testing.T) {
 				"a.go":      "package p; var A int",
 				"x_test.go": `package p_test; import "test/pkg"; var X = p.A`,
 				"y_test.go": "package p_test; func Y() int { return X }",
+
+				// non xtest to ensure we don't mix up xtest and test.
+				"a_test.go": `package p; var X = A`,
+				"b_test.go": "package p; func Y() int { return X }",
 			},
 			cases: lspTestCases{
 				wantHover: map[string]string{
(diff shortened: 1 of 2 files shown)


keegancsmith:
Funnily enough, a related commit I merged straight into master is broken with out this change. The test case used to pass due to the bug (we would have pkgs like `test/pkg/_test` as a package loaded, and would manage to find xtest deps). https://travis-ci.org/sourcegraph/go-langserver/builds/199618328

langserver/langserver_test.go
```diff
@@ -157,6 +157,7 @@ func TestServer(t *testing.T) {
 			fs: map[string]string{
 				"a.go":      "package p; var A int",
 				"a_test.go": `package p_test; import "test/pkg"; var X = p.A`,
```
sqs: We should have a test case or another entry here that is `a2_test.go` `package p; var X = whatever` to make sure that packages `a` and `a_test` are kept separate.

If we have or add such a test case and it passes, then LGTM.

keegancsmith: Just pushed the commits which do this. We do isolate the two definitions :)

When it comes to actual `UsesOf`, we don't do any normalization on the stuff we get back from loader, so it should be very accurate. The stripping of `_test` we do outside of that is purely heuristic based to help decide which packages deserve having UsesOf run on them.

PR #141, opened 2017-02-08
fs: Support incremental file updates

keegancsmith:
We still don't advertise we support this since I have not tested this
thoroughly. However, I am working on a PR now which refactors this logic to make
it more testable.


keegancsmith:
git is acting screwy

PR #142, opened 2017-02-08
fs: Support incremental file updates

keegancsmith:
We still don't advertise we support this since I have not tested this
thoroughly. However, I am working on a PR now which refactors this logic to make
it more testable.

### DIFF
--- a/langserver/fs.go
+++ b/langserver/fs.go
@@ -1,6 +1,7 @@
 package langserver
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


keegancsmith:
CI is acting very funny with this. Both runs should be testing identical commits, but they are not. The PR CI build is testing an older commit I pushed, and it won't update to the new one. CI is passing on this if travis tested the proper commit. It also passes locally.

sqs:
@keegancsmith Indeed, Travis CI does seem to be testing the wrong commit. Strange.

LGTM.

keegancsmith:
I am rebasing on latest master and repushing just to try and trick travis into giving me the green I deserve :)

PR #145, opened 2017-02-09
fs: Only invalidate cache if files change

keegancsmith:
See commit messages.

### DIFF
--- a/langserver/fs.go
+++ b/langserver/fs.go
@@ -37,8 +37,6 @@ func (h *HandlerShared) HandleFileSystemRequest(ctx context.Context, req *jsonrp
 	var before []byte
 	var beforeErr error
 	checkChanged := func(uri string) bool {
-		// We ignore readFile errors, since the []byte being nil gives
-		// us the same comparison.
 		b, err := h.readFile(ctx, uri)
 		if os.IsNotExist(beforeErr) && os.IsNotExist(err) {
 			// Other error conditions we are conservative and
--- a/langserver/fs.go
+++ b/langserver/fs.go
@@ -26,9 +26,9 @@ func IsFileSystemRequest(method string) bool {
 		method == "textDocument/didSave"
 }
 
-// HandleFileSystemRequest handles textDocument/did* requests. true is
-// returned if a file was modified.
-func (h *HandlerShared) HandleFileSystemRequest(ctx context.Context, req *jsonrpc2.Request) (bool, error) {
+// HandleFileSystemRequest handles textDocument/did* requests. The path the
+// request is for is returned. true is returned if a file was modified.
+func (h *HandlerShared) HandleFileSystemRequest(ctx context.Context, req *jsonrpc2.Request) (string, bool, error) {
 	span := opentracing.SpanFromContext(ctx)
 	h.Mu.Lock()
 	overlay := h.overlay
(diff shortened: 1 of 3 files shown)


langserver/fs.go
```diff
@@ -26,42 +26,59 @@ func IsFileSystemRequest(method string) bool {
...
 	case "textDocument/didOpen":
 		var params lsp.DidOpenTextDocumentParams
 		if err := json.Unmarshal(*req.Params, &params); err != nil {
-			return err
+			return "", false, err
 		}
-		span.SetTag("uri", params.TextDocument.URI)
+		uri := params.TextDocument.URI
+		span.SetTag("uri", uri)
+		before, _ = h.readFile(ctx, uri)
```
nicksnyder: Is it correct to ignore the error here? What if

1. This call fails (before is now nil)
2. The operation succeeds (in this case didOpen) and modifies the document
3. The h.readFile in checkChanged may also fail and conclude that nil == nil so no change happened?

Think of all the permutations of readFile failing (or not) and the operation modifying the document (or not).

If either of the readFile's fail I don't think we can conclude whether the file changed or not.

Looking at the call site, it seems like it would be prudent to assume that the file changed if either of the readFile calls fail.

keegancsmith: I was going from the assumption that an error means file not found => empty []byte is file not found, which makes the bytes.Equal check I do correct. However, there are (rarer) cases were we can get an error for a different reason. I'll correctly handle that.

keegancsmith:
@nicksnyder ping

langserver/fs.go
```diff
@@ -26,42 +26,67 @@ func IsFileSystemRequest(method string) bool {
...
 	span := opentracing.SpanFromContext(ctx)
 	h.Mu.Lock()
 	overlay := h.overlay
 	h.Mu.Unlock()
 
+	var before []byte
+	var beforeErr error
+	checkChanged := func(uri string) bool {
+		// We ignore readFile errors, since the []byte being nil gives
+		// us the same comparison.
```
nicksnyder: This comment seems out of date. Perhaps just document both the if and else cases below.

langserver/fs.go
```diff
@@ -26,42 +26,67 @@ func IsFileSystemRequest(method string) bool {
...
+// HandleFileSystemRequest handles textDocument/did* requests. The path the
+// request is for is returned. true is returned if a file was modified.
+func (h *HandlerShared) HandleFileSystemRequest(ctx context.Context, req *jsonrpc2.Request) (string, bool, error) {
 	span := opentracing.SpanFromContext(ctx)
 	h.Mu.Lock()
 	overlay := h.overlay
 	h.Mu.Unlock()
 
+	var before []byte
+	var beforeErr error
```
nicksnyder: It seems less than ideal that this state needs to get set before calling checkChanged.

I would recommend either making these parameters to checkChanged or better yet, making checkChanged take a URI and an operation closure and having checkChanged call readFile before/after the operation.

keegancsmith: Yeah I agree. I considered the closure option but thought it would be a perf overhead. I realise that is foolish now, will update :)

langserver/fs.go
keegancsmith: will fix in my update to use a closure instead.

langserver/fs.go
keegancsmith: I don't like how it is looking tbh. I'll push the change and tell me what you think.

keegancsmith: Actually it ended up alright

langserver/fs.go
```diff
@@ -26,42 +26,62 @@ func IsFileSystemRequest(method string) bool {
...
 	h.Mu.Unlock()
 
+	do := func(uri string, op func() error) (string, bool, error) {
+		span.SetTag("uri", uri)
+		before, beforeErr := h.readFile(ctx, uri)
+		err := op()
+		after, afterErr := h.readFile(ctx, uri)
+		if os.IsNotExist(beforeErr) && os.IsNotExist(afterErr) {
+			// Other error conditions we are conservative and
+			// assume something changed.
```
nicksnyder: This comment could just be `// File didn't exist before or after so nothing has changed`

Then in the else block you can say

```
// If an error prevented us from reading the file before or after
// then we assume the file changed to be conservative.
```

keegancsmith:
Thanks for the suggestions!

PR #149, opened 2017-02-13
references: Stream if streaming client capability is present

keegancsmith:
See https://github.com/sourcegraph/language-server-protocol/pull/19

### DIFF
--- a/pkg/lsp/service.go
+++ b/pkg/lsp/service.go
@@ -18,6 +18,16 @@ type ClientCapabilities struct {
 	// Below are Sourcegraph extensions. They do not live in lspext since
 	// they are extending the field InitializeParams.Capabilities
 
+	// Streaming indicates the client supports receiving the result solely
+	// through $/partialResult notifications for requests from the client
+	// to the server.
+	//
+	// If true, the server is allowed to let the result in the final
+	// response be null.
+	//
+	// Currently a soon to be proposed extension https://github.com/sourcegraph/language-server-protocol/pull/19
+	Streaming bool `json:"streaming,omitempty"`
+
 	// XFilesProvider indicates the client provides support for
 	// workspace/xfiles. This is a Sourcegraph extension.
 	XFilesProvider bool `json:"xfilesProvider,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/lsp/service.go
```diff
@@ -18,6 +18,16 @@ type ClientCapabilities struct {
 	// Below are Sourcegraph extensions. They do not live in lspext since
 	// they are extending the field InitializeParams.Capabilities
 
+	// Streaming indicates the client supports receiving the result solely
+	// through $/partialResult notifications for requests from the client
+	// to the server.
+	//
+	// If true, the server is allowed to let the result in the final
```
nicksnyder: FYI, I added a comment on the LSP proposal to update the language here.

pkg/lsp/service.go
```diff
@@ -18,6 +18,16 @@ type ClientCapabilities struct {
...
+	// response be null.
+	//
+	// Currently a soon to be proposed extension https://github.com/sourcegraph/language-server-protocol/pull/19
+	Streaming bool `json:"streaming,omitempty"`
```
felixfbecker: we should prefix this with `x` as long as our implementation is not accepted

keegancsmith:
@nicksnyder I haven't updated the language yet, I commented why here https://github.com/sourcegraph/language-server-protocol/pull/19/files#r101497385

@felixfbecker updated to be xstreaming.

keegancsmith:
@nicksnyder, @felixfbecker how should we proceed? I am happy to just close this if we aren't sure how to proceed.

felixfbecker:
Is the issue that the Go LS only streaming for a hand-selected methods?

keegancsmith:
Nope, we had all that discussion about what the streaming experiment should actually be which I don't think is resolved yet?

Go LS server as it stands only streams for references. It only does that if STREAM_EXPERIMENT is true. This PR also streams if the client advertises streaming in its caps. Otherwise this PR is just updating the relevant json structs to match what is in the spec.

felixfbecker:
The reason I asked is that the comment you linked was about the capability meaning that the server needs to stream all methods

keegancsmith:
Sure. Well I don't want to implement that in this PR. So given that if we use the capabilities that is what we want it to do, I'll just abandon this PR.

Issue #146, opened 2017-02-11
textDocument/references unexpected panic: sync: unlock of unlocked mutex


nicksnyder:
```
[...]

PR #151, opened 2017-02-14
references: Guard access to import graph once

keegancsmith:
I think import graph being a value + reset would lead to us overwriting the
value of the mutex => once trying to unlock a new mutex after it has finished
importing the graph.

Fixes https://github.com/sourcegraph/go-langserver/issues/146

### DIFF
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -37,7 +37,7 @@ type LangHandler struct {
 	symbolCache    cache
 
 	// cache the reverse import graph
-	importGraphOnce sync.Once
+	importGraphOnce *sync.Once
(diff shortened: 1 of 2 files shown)


langserver/handler.go
```diff
@@ -37,7 +37,7 @@ type LangHandler struct {
 	symbolCache    cache
 
 	// cache the reverse import graph
-	importGraphOnce sync.Once
+	importGraphOnce *sync.Once
```
nicksnyder: Please add a comment to explain why this is a pointer and not a value.

It would also be nice to add a comment to mu to describe what it does and does not protect.

PR #153, opened 2017-02-20
Fixes #136 Use position to determine activeparam for signature help

### LINKED ISSUE #136: Active Parameter in Signature Help gets incremented before typing comma
I was testing the integration of the language server with VS Code.

Take the example of `math.max()`. This takes 2 parameters
When cursor is inside the brackets, active parameter is shown as the first one. 

screen shot 2017-02-05 at 6 31 24 pm

Type a single digit, the second parameter becomes the active parameter.

screen shot 2017-02-05 at 6 31 39 pm

Expected: The second parameter becomes the active one only after pressing comma

ramya-rao-a:
cc @keegancsmith 

Fixes #136


langserver/signature.go
```diff
@@ -37,6 +37,28 @@ func (h *LangHandler) handleTextDocumentSignatureHelp(ctx context.Context, conn
 	if activeParameter > 0 {
 		activeParameter = activeParameter - 1
 	}
+
+	filename := h.FilePath(params.TextDocument.URI)
```
ramya-rao-a: All this is already done in `typecheck()`. 
We can just return the `start` value from `typecheck()`
But before I went about doing that I wanted to know if there is any simpler way of getting the `Pos` equivalent of given `lsp.Position` ?

langserver/signature.go
```diff
@@ -37,6 +37,28 @@ func (h *LangHandler) handleTextDocumentSignatureHelp(ctx context.Context, conn
...
+	}
+	offset, valid, _ := offsetForPosition(contents, params.Position)
+	if !valid {
+		return nil, nil
+	}
+	start := posForFileOffset(fset, filename, offset)
+	if start == token.NoPos {
+		return nil, nil
+	}
+
```
ramya-rao-a: All of the above is already done in `typecheck()`. 
We can just return the `start` value from `typecheck()`
But before I went about doing that I wanted to know if there is any simpler way of getting the `Pos` equivalent of given `lsp.Position` ?

ramya-rao-a:
Am not sure why the test is failing, @keegancsmith can you help?

keegancsmith:
The test that is failing is related to the signature code.
```
package p; func main() { B(); A(); A(0,) }
                                       ^
```
The old code would say you are asking about the 2nd parameter. After your change we say you are asking about the first parameter. Hence the failing test case.

ramya-rao-a:
@keegancsmith Turns out the test case was using wrong position. It was using 1-based line number but 0 based character/column number. The tests passed due to the bug I am trying to fix :)

Updated the column number to be 1-based and added more test cases.

ramya-rao-a:
@keegancsmith `posForFileOffset` needs `offset` which you can get from `offsetForPosition` which needs file contents for which I would have to do a `readFile`
All that is already done in `typecheck`, so why do all that one more time?

Rebase is done

PR #158, opened 2017-02-21
references: Include declaration in stream

### LINKED ISSUE #157: Last reference not sent via $/partialResult
I am the streaming references experiment enabled locally and it seems like the last reference is never sent via $/partialResult.

For example, if I expect 4 references back, I will get 3 references from $/partialResult and the final response will return all 4 references. This happens even if I delay the final response by an arbitrary amount.

Given the streaming semantics we discussed, it should be possible to construct the entire result from all $/partialResults.

keegancsmith:
Fixes #157

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -105,6 +105,11 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 		close(locsC)
 	}()
 
+	// Don't include decl if it is outside of workspace.
+	if params.Context.IncludeDeclaration && PathHasPrefix(defpkg, h.init.RootImportPath) {
+		refs <- &ast.Ident{NamePos: obj.Pos(), Name: obj.Name()}
+	}
+
 	// seen keeps track of already findReferenced packages. This allows us
 	// to avoid doing extra work when we receive a successive import
 	// graph.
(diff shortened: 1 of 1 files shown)


langserver/references.go
```diff
@@ -105,6 +105,11 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 		close(locsC)
 	}()
 
+	// Don't include decl if it is outside of workspace.
+	if params.Context.IncludeDeclaration && PathHasPrefix(defpkg, h.init.RootImportPath) {
```
nicksnyder: I think this needs to happen before (or in) the goroutine. Otherwise it is racing with the close on the channel.

Since the declaration requires no computation, it seems like it should be first anyway?

keegancsmith: only this goroutine sends to the chan. Also the chan is not buffered, so this will block anyways until the collector collects it. I'll update to send this ref straight away.

keegancsmith:
I'll add some more comments unrelated to this PR so it is clearer who owns what.

langserver/references.go
nicksnyder: Ah, got it. I confused locsC with refs.

PR #164, opened 2017-02-24
Remove sorting

nicksnyder:
Sorting should be done by client


slimsag:
Does VS Code handle sorting on it's own, too? In the case of go-langserver, because results are grabbed in parallel everywhere, it'd be very bad if the results were not sorted before displaying to a user.

what is the motivation for such a change?

nicksnyder:
> Does VS Code handle sorting on it's own

Yes.

> In the case of go-langserver, because results are grabbed in parallel everywhere, it'd be very bad if the results were not sorted before displaying to a user.

In general, it is the UI's job to present results to the user in the best way (including sorting). Server doesn't need to sort.

> what is the motivation for such a change?

With streaming results, sorting on the server is irrelevant anyway (since results will be returned as they are found) and clients should not rely on the server to sort results.

This is just extra work that the server doesn't need to do and shouldn't do.

nicksnyder:
Also, LSP doesn't mention sorting anywhere so clients can't rely on language servers sorting results, which means they have to do it on the client anyway.

slimsag:
Thanks for the explanations :) LGTM

PR #169, opened 2017-03-01
hover: Include struct and interface fields

### LINKED ISSUE #168: Struct members are not seen on hover
See https://github.com/Microsoft/vscode-go/issues/706#issuecomment-283220940

keegancsmith:
For structs and interfaces we now include an extra hover content field containing the fields in the struct.

image

image

Fixes #168

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -55,6 +55,7 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 	qf := func(*types.Package) string { return "" }
 
 	var s string
+	var extra string
 	if f, ok := o.(*types.Var); ok && f.IsField() {
 		// TODO(sqs): make this be like (T).F not "struct field F string".
 		s = "struct " + o.String()
@@ -63,6 +64,11 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 			typ := obj.Type().Underlying()
 			if _, ok := typ.(*types.Struct); ok {
 				s = "type " + obj.Name() + " struct"
+				extra = prettyPrintTypesString(types.TypeString(typ, qf))
+			}
+			if _, ok := typ.(*types.Interface); ok {
+				s = "type " + obj.Name() + " interface"
+				extra = prettyPrintTypesString(types.TypeString(typ, qf))
 			}
 		}
 		if s == "" {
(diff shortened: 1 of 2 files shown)


junkblocker:
The original display shows raw strings as it is while sourcegraph display shows them as cooked strings. See:

Original version:
https://gyazo.com/f16dc38308c76fe0997a0172f485bc09

Sourcegraph version:
https://gyazo.com/82df5716b5fb15e0bcbbb8f0af417b32

Though it is cool that the Sourcegraph version also shows godoc.

keegancsmith:
@junkblocker shouldn't you just peak definition for that?

junkblocker:
@keegancsmith, yes but that's irrelevant to comparing current vs new implementation hovertip views.

keegancsmith:
@ramya-rao-a is the goal for us to be exactly the same as the older version? I personally prefer the new implementation for hover vs what vscode-go used to do (in fact I also want to get rid of the tags). But I'm happy to make it more like the old implementation if that is what makes sense.

Also in future, if we get the signature helper to work on structs I suppose that would help the common reason people look at this info?

grobie:
I also prefer the new implementation.

junkblocker:
@keegancsmith , @ramya-rao-a - Can't this be a switch? I'd prefer to see both the uncooked display and the struct tags.

ramya-rao-a:
There can always be cases where a feature implementation by the language server is different than that of what vscode-go has at the moment. It might also be better. So no, the goal is not to "exactly" match the existing behavior. But it should match existing needs.

In the current case, having the docs squeezed in between the struct type and struct members does look a bit odd, but that's just a personal preference.

I'd say, lets keep it as is, and wait for more feedback from users.

This would also affect hover in source graph, do you (at source graph) have any preference?

ramya-rao-a:
It would be nice to have a poll :)

sqs:
Making the struct tags raw strings is definitely better (`Foo string \`json:"foo"\`` instead of `Foo string "json:\"foo\""`). We'll definitely do that. I think that's a no-brainer.

As for showing the doc synopsis (first sentence), Sourcegraph users generally do like seeing that in addition to the struct definition. The improvement we should make is to display the doc prose in a sans-serif font, not in monospace. That'll better distinguish the doc synopsis from the struct definition, and it'll make the doc synopsis take up less space.

Feel free to upvote this if you like seeing doc synopses and downvote if you do not. :smile:

ramya-rao-a:
@sqs Just to be clear, when you say "doc synopsis" you mean the first line and "doc prose" you mean the second correct?

image

sqs:
Actually, I meant those interchangeably, but I was mistaken. I thought we were truncating the doc string to the first sentence as is often done for Go (using https://golang.org/pkg/go/doc/#Synopsis). Apparently we are not anymore. My bad.

Anyway, my proposal was to show a sans-serif first sentence docstring only, plus the (monospace) type/struct definition. This way, you get most of the benefit of having the docstring without it taking up a lot of space. Seems better than no-docstring or full-docstring to me, and it follows Go convention.

But very happy to change it or allow configuration if people don't like it this way.

ramya-rao-a:
We should definitely include the doc string. No arguments there.

Regarding the font issue, what is happening is the language server is returning a markdown content for the hover and not simple text. Simple text is shown in a smaller font (you can see the current behavior in VS Code), markdown for some reason appears bigger and different font. I'll take a look at that separately from VS Code side.

Things left to debate/discuss/vote/poll are

- Show raw original strings vs cooked strings for the members
- Include the members in the first block right with the type definition followed by the doc string in the second block (current behavior in VS Code) or the members appear after the doc string in the third block (behavior shown in the description of this issue)

Issue #165, opened 2017-02-28
Enable & fix limit parameter to */references requests


slimsag:
- `textDocument/references` doesn't respect the limit parameter when streaming, only when the final results are returned.
- `workspace/references` needs a limit parameter, and should respect it for the final results returned & during streaming.

keegancsmith:
I assume we should make the limit much higher than 100 then. Also do we have any concerns around quality of results when we start to limit. Previously we sorted in the langserver and then limited results, which would result in more relevant results within the limit.

Either way this should be easy to implement. Do we have any idea of how soon we want this in production?

slimsag:
From my discussion with @nicksnyder -- we settled on him using the specs I wrote above _and_ limiting the results on the frontend temporarily (b/c Go/Java LS do not respect it). So nothing is blocked on this / not a super high priority, but good to implement + spec out properly sometime soon.

nicksnyder:
> Previously we sorted in the langserver and then limited results, which would result in more relevant results within the limit.

This doesn't help in the context of streaming though. Client does its own sorting anyway.

> Also do we have any concerns around quality of results when we start to limit.

Yes and no. To the extent that the language server can stream the most relevant results first (e.g. walk "close" files/dirs first) then it should. Other then that, the UI is just going to take the first N results it and filter. We can adjust N if we feel like results aren't relevant enough.

Since the frontend wants to limit anyway, the backend might as well not waste the time/resources sending the rest of the results.

keegancsmith:
I have added https://github.com/sourcegraph/go-langserver/pull/173 which adds a limit param to xrefs. I think once that lands, we can close this out. However, we still need to update the spec and use the limit parameter in the frontend. + We should remove the 15s timeout once both of those have been completed.

PR #171, opened 2017-03-01
references: Respect limit when streaming

keegancsmith:
Part of #165

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -76,6 +76,12 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 		return PathHasPrefix(path, h.init.RootImportPath)
 	}
 
+	// findRefCtx is used in the findReferences function. It has its own
+	// context so we can stop finding references once we have reached our
+	// limit.
+	findRefCtx, stop := context.WithCancel(ctx)
+	defer stop()
+
 	var (
 		// locsC receives the final collected references via
 		// refStreamAndCollect.
@@ -93,7 +99,7 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 	// references back to the client, as well as build up the final slice
 	// which we return as the response.
 	go func() {
-		locsC <- refStreamAndCollect(ctx, conn, req, fset, refs, params.Context.XLimit)
+		locsC <- refStreamAndCollect(ctx, conn, req, fset, refs, params.Context.XLimit, stop)
 		close(locsC)
 	}()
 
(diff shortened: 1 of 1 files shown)


langserver/references.go
```diff
@@ -75,6 +74,12 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 		return PathHasPrefix(path, h.init.RootImportPath)
 	}
 
+	// findRefCtx is used in the findReferences function. It has its own
+	// context so we can stop finding references once we have reached our
+	// limit.
+	findRefCtx, stop := context.WithCancel(ctx)
```
nicksnyder: just curious, is there a reason why we wouldn't want to just shadow ctx at the beginning of the function to avoid accidentally sending the wrong ctx later?

langserver/references.go
```diff
@@ -92,7 +97,7 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 	// references back to the client, as well as build up the final slice
 	// which we return as the response.
 	go func() {
-		locsC <- refStreamAndCollect(ctx, conn, req, fset, refs)
+		locsC <- refStreamAndCollect(ctx, conn, req, fset, refs, params.Context.XLimit, stop)
```
slimsag: Do we need `findRefCtx` here instead?

langserver/references.go
```diff
@@ -144,34 +149,21 @@ func (h *LangHandler) handleTextDocumentReferences(ctx context.Context, conn jso
 			lconf.ImportWithTests(path)
 		}
 
-		findRefErr = findReferences(ctx, lconf, pkgInWorkspace, obj, refs)
+		findRefErr = findReferences(findRefCtx, lconf, pkgInWorkspace, obj, refs)
```
nicksnyder: should we stop iterating if the err is due to cancel?

langserver/references.go
```diff
@@ -296,6 +296,10 @@ func refStreamAndCollect(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonr
 				send()
 				return locs
 			}
+			if len(locs) >= limit {
+				stop()
```
nicksnyder: just to make sure I read this correct, this limit parameter is just best effort. There may be more than limit number of results returned?

Is there a reason you chose this over limiting the results on the producer side of the refs channel?

langserver/references.go
keegancsmith: we only want to cancel finding references. There are other operations happening, but it would actually work if we just shadowed at the beginning. However, in reality all we want to do is cancel reference finding, so I believe this more accurately shows the intent + future proof for extra things we may add.

langserver/references.go
keegancsmith: no

langserver/references.go
keegancsmith: I'll add that

langserver/references.go
keegancsmith: > just to make sure I read this correct, this limit parameter is just best effort. There may be more than limit number of results returned?

No. We will not send more than limit. This is the goroutine that is responsible for collecting references and sending them. When it calls stop, it is telling the producer it can stop searching for references. If the continue statement just below this line was not there, we could potentially send more found references.

> Is there a reason you chose this over limiting the results on the producer side of the refs channel?

there is one place we consume from the channel, but more than one place we produce. So it is cleaner this way.

PR #172, opened 2017-03-06
Handle didChange requests in order

### LINKED ISSUE #155: Enforce LSP ordering semantics
LSP requires an ordering on how the server handles requests:
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#request-notification-and-response-ordering

If a client sends LSPs request in rapid succession (e.g. textDocument/didOpen, textDocument/didChange, textDocument/didClose), the server MUST handle these events in the order they are received.

JSON RPC does not have any ordering requirements so JSON RPC libraries are not required to provide these ordering semantics.

For example, if each notification handler spawns a new goroutine, there is no guarantee of the order that requests will be handled.

The Go language server needs to implement these ordering semantics.

keegancsmith:
Currently we use an AsyncHandler, which means we could process didChange
requests out of order, depending on how the go runtime decides to schedule the
goroutines.

Fixes #155

### DIFF
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -21,9 +21,33 @@ import (
 
 // NewHandler creates a Go language server handler.
 func NewHandler() jsonrpc2.Handler {
-	return jsonrpc2.AsyncHandler(jsonrpc2.HandlerWithError((&LangHandler{
+	return lspHandler{jsonrpc2.HandlerWithError((&LangHandler{
 		HandlerShared: &HandlerShared{},
-	}).handle))
+	}).handle)}
+}
+
+// lspHandler wraps LangHandler to correctly handle requests in the correct
+// order.
+//
+// The LSP spec dictates a strict ordering that requests should only be
+// processed serially in the order they are received. However, implementations
+// are allowed to do concurrent computation if it doesn't affect the
+// result. We actually can return responses out of order, since vscode does
+// not seem to have issues with that. We also do everything concurrently,
+// except methods which could mutate the state used by our typecheckers (ie
+// textDocument/didOpen, etc). Those are done serially since applying them out
+// of order could result in a different textDocument.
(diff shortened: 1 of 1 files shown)


keegancsmith:
Note: We will need to update the buildserver on sourcegraph to use the correct behaviour (similiar change). This is a fix for the langserver used in vscode-go.

langserver/handler.go
```diff
@@ -21,9 +21,33 @@ import (
...
+// except methods which could mutate the state used by our typecheckers (ie
+// textDocument/didOpen, etc). Those are done serially since applying them out
+// of order could result in a different textDocument.
+type lspHandler struct {
+	jsonrpc2.Handler
+}
+
+// Handle implements jsonrpc2.Handler
+func (h lspHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jsonrpc2.Request) {
+	if isFileSystemRequest(req.Method) {
```
nicksnyder: is this sufficient?

What if textDocument/references or other non-file system requests are processed before textDocument/didOpen or after textDocument/didClose?

keegancsmith: It kind of acts like changes to the filesystem at the moment, it will see whatever the state of the FS at the time we do read. However, this guarantees exactly what we need to give the correct behaviour to the end user:

* incremental FS mutations are processed in the correct order.
* a reference request after a FS mutation will see the at least that version (same as if it was backed by a mutable FS)

Technically an older request that is still running may see state from a later FS mutation. However, I believe this is fine since that is what is in the users editor + vscode has likely already canceled the older request.

Making it so that non-mutable requests have a snapshot of the overlay FS is possible to give the full guarantee. However, I don't believe it is necessary + is more complicated + is impossible if it was backed by an FS anyways.

PR #173, opened 2017-03-06
Add Limit param to workspace/xreferences

keegancsmith:
Part of #165


pkg/lspext/lspext.go
```diff
@@ -27,6 +27,9 @@ type WorkspaceReferencesParams struct {
 	// look in order to find the symbol (this is an optimization). It is up to
 	// the language server to define the schema of this object.
 	Hints map[string]interface{} `json:"hints,omitempty"`
+
+	// Limit if positive will limit the number of results returned.
+	Limit int `json:"xlimit,omitempty"`
```
slimsag: copy+paste error here, `xlimit` should be `limit`

keegancsmith: 🤦‍♂️

PR #181, opened 2017-04-03
make `MarkedString` more spec compliant

beyang:
The LSP spec defines `MarkedString` as follows:

```
type MarkedString = string | { language: string; value: string };
```

This makes MarkedString serializable/deserializable to raw strings and adds unit tests.

### DIFF
--- a/pkg/lsp/service.go
+++ b/pkg/lsp/service.go
@@ -3,6 +3,7 @@ package lsp
 import (
 	"bytes"
 	"encoding/json"
+	"strings"
 )
 
 type None struct{}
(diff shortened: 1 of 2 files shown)


pkg/lsp/service.go
```diff
@@ -200,9 +201,36 @@ type Hover struct {
 	Range    *Range         `json:"range,omitempty"`
 }
 
-type MarkedString struct {
+type MarkedString markedString
+
+type markedString struct {
```
keegancsmith: I don't understand the reason for the type alias? You should just be able to implement the marshallers on `type MarkedString struct`

beyang: I wanted to take advantage of the default behavior of `json.Unmarshal` (less code to update if we later add more fields to the struct). This is the best I could come up with, but happy to amend if there's a better way.

ramya-rao-a:
Thanks @beyang !

This lets me fix the hover issue: https://github.com/Microsoft/vscode-go/issues/852

PR: #182

PR #182, opened 2017-04-04
Send comments in hover as string

ramya-rao-a:
This PR is to fix https://github.com/Microsoft/vscode-go/issues/852

@keegancsmith The tests at https://github.com/sourcegraph/go-langserver/blob/master/langserver/langserver_test.go#L778 are failing.

There is an expected ` \n\n` but it is finding only `\n`

My guess is that this is due to trimming at https://github.com/sourcegraph/go-langserver/blob/1212ac52443c77f5e6530e9e11f5793b3e9de0fe/pkg/lsp/service.go#L214

I am not clear why we are adding the extra `\n` in the first place.

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -3,6 +3,7 @@ package langserver
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"go/ast"
 	"go/build"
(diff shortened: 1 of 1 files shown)


ramya-rao-a:
Closing in favor of #183

PR #184, opened 2017-04-18
do not emit a containerName in the SymbolInformation for package-scope defs

sqs:
vscode's internal `skipLocalSymbols` option treats SymbolInformations with containerName set as locals, which are skipped. This means that when that option is used, there are no Go symbol results.

This fixes our interpretation of containerName to match vscode's assumptions.

There is minimal end-user impact. If there is no containerName, vscode shows the file path, which is actually more useful than the containerName we previously used (the package name).

### DIFF
--- a/langserver/langserver_test.go
+++ b/langserver/langserver_test.go
@@ -73,43 +73,43 @@ func TestServer(t *testing.T) {
 					},
 				},
 				wantSymbols: map[string][]string{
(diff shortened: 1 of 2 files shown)


keegancsmith:
Could this break assumptions in our global symbol indexing on sourcegraph.com? cc @slimsag

PR #189, opened 2017-05-01
refs: Remove hardcoded timeout from workspace/refs

keegancsmith:
This is a relic from before we had streaming support. We still don't support
passing on `$/cancelRequest`. However, that should be fine since it seems to be
fine for references.


nicksnyder:
I think it would be better to keep a finite timeout but just increase it to something larger like 1 minute.

PR #191, opened 2017-05-16
Always send $/partialResult

keegancsmith:
If a client does not support $/partialResult, it will just ignore the
notfications.


nicksnyder:
Remember to remove STREAM_EXPERIMENT from sourcegraph/infrastructure once this is merged/deployed

Issue #178, opened 2017-03-14
High memory & CPU usage after running for awhile


beyang:
Enabled the langserver and did a few hovers for testing. Then vscode was in the background. After a few minutes the langserver was using 5.6GB of memory (machine only has 4 GB RAM).

keegancsmith:
Related to https://github.com/Microsoft/vscode-go/issues/853

So the issue is our language server resource usage is currently tuned for a server environment were we have a premise of lots of RAM. This has been presenting us issues in prod as well, so rewriting our hover and j2d to be smarter with resources needs to be done. It is almost certainly related to using the simple go loader interface and/or naive caching of its output. Add in our conservative cache invalidation policy, and we create a lot of garbage and re-work things out often in an editor environment.

I'll see if there is any short term wins here, but we probably need to make some larger changes to get good at this.

vanackere:
Same issue for me: even with 16GB of memory I had to stop using go-langserver in vscode because it was frequently using all memory, making the computer unusable and triggering the OOM-Killer...

vanackere:
Since I can reliably cause the langserver to use all memory on my private code base., how can I help debug this issue (or at least find where most of the ressource are spent) ? Any profiling tool to run ? If so, how ?

ramya-rao-a:
@keegancsmith Any updates on this?

I have pushed a change for adding the ability to pass `-trace` and `-logfile` flags to the language server (See https://github.com/Microsoft/vscode-go/commit/50e5c8acd0553e04572b221924149b1b2f7926df)

Is there anything else you'd like to see users able to do?

OneOfOne:
I hate to do this but any news on this? this is kind of a major issue.

I have 32gb of ram and the oom killer triggered twice in the last 4 hours, gocode is broken on go git.

beyang:
@OneOfOne Apologies for the delay. We're still looking into it, but have had to prioritize a few other features for Go in the past 2 weeks. I'll update this issue when we have a patch targeting this issue. Appreciate the feedback and thanks for trying it out!

mgleason:
+1 Same issue here (on MacOS).  Seems to grow to consume all available memory and then keeps going consuming swap as the system slows to a crawl.   

[...]

PR #192, opened 2017-05-31
significantly lower memory consumption for vscode-go users

slimsag:
- This change significantly lowers memory consumption when running in editor mode (i.e. with vscode-go). From 7GB down to 4GB for a large project such as Docker.
- Additionally adds a pprof HTTP server by default, for capturing profiles easily (see README).
- Also adds an `-freeosmemory=true` flag, which when set to false can be used to disable this behavior for testing purposes (we'll also probably disable on sourcegraph.com).

Significantly helps https://github.com/sourcegraph/go-langserver/issues/178 (probably fixes entirely, but I have more improvements coming soon that I believe will lower memory consumption even further :) ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,21 @@ go get github.com/sourcegraph/go-langserver
 |    | Hover | Jump to def | Find references | Workspace symbols | VFS extension | Isolated | Parallel |
 |----|-------|-------------|-----------------|-------------------|---------------|----------|----------|
 | Go |   x   |      x      |        x        |         x         |       x       |     x    |     x    |
+
+## Profiling
+
+If you run into performance issues while using the language server, it can be very helpful to attach a CPU or memory profile with the issue report. To capture one, first [install Go](https://golang.org/doc/install) and then:
+
+Capture a heap (memory) profile:
+
+```bash
+go tool pprof -svg $GOPATH/bin/go-langserver http://localhost:6060/debug/pprof/heap > heap.svg
+```
+
+Capture a CPU profile:
+
+```bash
+go tool pprof -svg $GOPATH/bin/go-langserver http://localhost:6060/debug/pprof/profile > cpu.svg
+```
+
+Since these capture the active resource usage, it's best to run these commands while the issue is occuring (i.e. while memory or CPU is high).
--- a/main.go
+++ b/main.go
@@ -9,6 +9,8 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"runtime/debug"
+	"time"
 
 	"github.com/sourcegraph/go-langserver/langserver"
 	"github.com/sourcegraph/jsonrpc2"
@@ -23,6 +25,7 @@ var (
 	logfile      = flag.String("logfile", "", "also log to this file (in addition to stderr)")
 	printVersion = flag.Bool("version", false, "print version and exit")
 	pprof        = flag.String("pprof", ":6060", "start a pprof http server (https://golang.org/pkg/net/http/pprof/)")
+	freeosmemory = flag.Bool("freeosmemory", true, "aggressively free memory back to the OS")
 )
 
 // version is the version field we report back. If you are releasing a new version:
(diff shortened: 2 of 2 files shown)


langserver/loader.go
```diff
@@ -306,3 +312,47 @@ func isMultiplePackageError(err error) bool {
...
+// 	|-------------|------------|-------------|-----------|----------|-----------|
+// 	| 7.61GB      | 4.12GB     | -45.86%     | 3.92GB    | 3.33GB   | -15.05%   |
+//
+// Where `Real` means real memory reported by OS X Activity Monitor, and `Go`
+// means memory reported by Go as being in use.
+//
+// TL;DR: 46% less memory consumption for users running with the vscode-go extension.
+//
+// [1] https://github.com/golang/go/issues/14735#issuecomment-194470114
+func freeOSMemory() {
```
sqs: This feels like something that should be done in main.go, not in the library code. Also, then you wouldn't need to pass the `editor bool` all the way down, and you wouldn't need a `sync.Once`.

langserver/loader.go
```diff
@@ -306,3 +312,47 @@ func isMultiplePackageError(err error) bool {
...
+// quite a lot because we perform so many short-burst heap allocations during
+// the type-checking phase.
+//
+// This function should only be invoked in editor mode, not in sourcegraph.com
+// mode, because users running the language server as part of their editor
+// generally expect much lower memory usage. In contrast, on sourcegraph.com we
+// can give our servers plenty of RAM and allow Go to consume as much as it
+// wants. Go does reuse the memory not free'd to the OS, and as such enabling
+// this does _technically_ make our application perform less optimally -- but
+// in practice this has no observable affect in editor mode.
```
sqs: typo: affect -> effect

langserver/loader.go
slimsag: Yep, that's true. The reason I did it this way is to keep it close to the problem (type-checking) and because I plan to use the `editor` bool in this same location in another change soon.

langserver/loader.go
```diff
@@ -306,3 +312,47 @@ func isMultiplePackageError(err error) bool {
...
+//
+// The end effect of performing this is that repeating "hover over code" -> "make an edit"
+// 10 times inside a large package like github.com/docker/docker/cmd/dockerd:
+//
+//
+// 	| Real Before | Real After | Real Change | Go Before | Go After | Go Change |
+// 	|-------------|------------|-------------|-----------|----------|-----------|
+// 	| 7.61GB      | 4.12GB     | -45.86%     | 3.92GB    | 3.33GB   | -15.05%   |
```
sqs: After this change, does the real memory usage continue to grow, just at a slower rate? Or does it plateau at some amount (4.12 GB here)?

felixfbecker:
I don't like the concept of an opt-in "editor mode" for language servers or anything that makes the language server behave differently for specific clients. Wouldn't a specific flag do the job too (to enable/disable the aggressive memory freeing, preferably with the default being enabled)?

langserver/loader.go
slimsag: Great question.

I have observed that both before this change and after this change, the memory does plateau at a certain amount (8.3GB before this change, and 7.22GB after this change) and does not continue to grow. That amount can just be quite large for projects with many dependencies.

The amount appears to be 2x the currently in-use Go heap size (~3.3GB), which makes sense. It's not clear to me why Go can only release 1GB of the ~4 GB difference in unused memory, but I speculate it could be due to issues like https://github.com/golang/go/issues/14045 or https://github.com/golang/go/issues/14735 since our allocations all come in large spikes due to type-checking.

If we can make type-checking faster more fine-grain caching (which I am doing), then we can potentially spare ourselves keeping a ~3.3GB Go heap in-memory all the time and lower this number even further (in addition to performing less type-checking on-the-fly).

slimsag:
@felixfbecker Understood. I have good reasons outside of this PR for why we might want such a mode, but let's punt on that discussion for now / wait until I have a provable need in a separate PR :) In the morning I'll update this PR to replace `-editor=true` with `-freeosmemory=true` (naming TBD) and will move the `FreeOSMemory` call into `main.go` per Quinn's suggestion

PR #196, opened 2017-06-23
langserver: add godef-based hover backend

slimsag:
This change adds a godef-based hover backend, similar to the godef-based
`textDocument/definition` backend added prior. The motivation for this
change is to avoid typechecking the entire program just to serve a single
`textDocument/hover` request. After this change, hover and definition are
both very quick and use little resources. Larger requests like `workspace/symbol`,
or `textDocument/references`, will continue to use more resources as they must
perform typechecking.

The output style of this hover implementation does vary from our prior
typechecking implementation in slight ways, but overall the implementation
always produces results that are on par or slightly better than our typechecking
implementation.

As with the `textDocument/definition` implementation, we should attempt consolidation
of this hover implementation with our typechecking variant further in the future. At
this point, of course, they are too different to share much code or implementation.

Helps https://github.com/sourcegraph/go-langserver/issues/178
Helps https://github.com/Microsoft/vscode-go/issues/853

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -6,8 +6,12 @@ import (
 	"fmt"
 	"go/ast"
 	"go/build"
+	"go/format"
+	"go/parser"
 	"go/token"
 	"go/types"
+	"path/filepath"
+	"sort"
 	"strings"
 
 	doc "github.com/slimsag/godocmd"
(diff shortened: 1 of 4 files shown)


langserver/hover.go
```diff
@@ -245,3 +253,272 @@ func prettyPrintTypesString(s string) string {
...
+
+	if loc.URI == "file://" {
+		// TODO: builtins do not have valid URIs or locations.
+		return &lsp.Hover{}, nil
+	}
+
+	filename := uriToFilePath(loc.URI)
+
+	// Parse the entire dir into its respective AST packages.
+	pkgs, err := parser.ParseDir(fset, filepath.Dir(filename), nil, parser.ParseComments)
```
keegancsmith: we have ways to parse that use the virtual filesystem. If you use that you will correctly parse unsaved changes. But it is likely a very minor issue, and I think we can fix all that when we consolidate the implementations.

slimsag: Thanks for the comment. This reminded me that I needed to fix hover & jump to definition for unsaved files.

I didn't address your comment directly, though, because as you can see in https://github.com/sourcegraph/go-langserver/pull/196/commits/8d908ebbf2ce2d3326e889e686890fd31c90fcbc the fix is not quite as simple. We can't use `parseDir` directly here because we need it to be aware of the test-suite somewhat. I agree, though, that this is something we can resolve down the line / is very minor.

PR #197, opened 2017-07-07
DONOTMERGE: Unify hover/definition (remove non-godef implementation)

slimsag:
Overview:

1. Remove unnecessary code / perform general cleanups.
2. Perform 100% of 'godef' hover/definition file IO through VFS. **Important: godef doesn't use .a files (and never has, I was wrong about this before).**
3. Unify our `textDocument/hover` and `textDocument/definition` endpoints: we used to have one of each based on godef (for editors), and another based on typechecking (for sourcegraph.com); we now just have the godef one for both cases. It's better all around, but especially in terms of performance.

Q&A

- Q: Why do some early commits reference a callback for 'go install'ing packages?
  - A: This was an early approach, but was removed later on (see more recent commits). I found out it wasn't needed, because godef doesn't use the `.a` files at all.
- Q: Does godef really not use .a files? What makes it fast?
  - A: Please see commit description https://github.com/sourcegraph/go-langserver/commit/c4e866ee63f1619581aa15b6fbb598023364d00b

Notes

- `textDocument/xdefinition` still uses typechecking, and can/should probably be updated to match this new approach.
- drawback: The godef-based hover implementation (which will be used on sourcegraph.com after this change) is based on `go/doc` which means it doesn't support e.g. hovering over a variable defined inside a function. We can fix this in the long run with a better pure AST implementation.

### DIFF
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -119,6 +119,11 @@ func (h *LangHandler) handle(ctx context.Context, conn *jsonrpc2.Conn, req *json
 	return h.Handle(ctx, conn, req)
 }
 
+// UsePremptiveTypechecking controls whether or not typechecking should be done
+// as early as possible. For just hover/definition, this does not improve time
+// to results, so it's advised for it to be off in local editor environments.
+var UsePremptiveTypechecking = false
+
 // Handle creates a response for a JSONRPC2 LSP request. Note: LSP has strict
 // ordering requirements, so this should not just be wrapped in an
 // jsonrpc2.AsyncHandler. Ensure you have the same ordering as used in the
@@ -356,7 +361,7 @@ func (h *LangHandler) Handle(ctx context.Context, conn jsonrpc2.JSONRPC2, req *j
 				// a user is viewing this path, hint to add it to the cache
 				// (unless we're primarily using binary package cache .a
 				// files).
-				if !UseBinaryPkgCache {
+				if !UsePremptiveTypechecking {
 					go h.typecheck(ctx, conn, uri, lsp.Position{})
 				}
 			}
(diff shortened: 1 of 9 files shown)


langserver/handler.go
```diff
@@ -119,6 +119,11 @@ func (h *LangHandler) handle(ctx context.Context, conn *jsonrpc2.Conn, req *json
 	return h.Handle(ctx, conn, req)
 }
 
+// UsePremptiveTypechecking controls whether or not typechecking should be done
+// as early as possible. For just hover/definition, this does not improve time
+// to results, so it's advised for it to be off in local editor environments.
+var UsePremptiveTypechecking = false
```
slimsag: I'll invert this boolean tomorrow

keegancsmith:
This is awesome btw! I'll get to reviewing it, but also will test to see how it works on Sourcegraph. I haven't reviewed it yet, but now that it is 100% VFS based can we still j2d/hover on things that are in the goroot or outside the workspace (but in GOPATH) for the local use case?

langserver/langserver_test.go
```diff
@@ -1111,15 +1111,18 @@ func lspTests(t testing.TB, ctx context.Context, fs *AtomicFS, c *jsonrpc2.Conn,
...
-		oldGOPATH := os.Getenv("GOPATH")
-		os.Setenv("GOPATH", tmpDir)
-		out, err := exec.Command("go", "install", "-v", "...").CombinedOutput()
-		os.Setenv("GOPATH", oldGOPATH)
-		if err != nil {
-			t.Fatal(err)
+		BuildGoPackage = func(path string) error {
+			// Install all Go packages in the $GOPATH.
+			oldGOPATH := os.Getenv("GOPATH")
+			os.Setenv("GOPATH", tmpDir)
```
keegancsmith: rather set this env on the exec.Command only. Otherwise this seems quite racy.

slimsag:
Closing for now, I do not know if/when I will get to this again. There are some major outstanding issues here.

PR #204, opened 2017-09-21
WIP documentSymbol supports receiver in different file to struct

### LINKED ISSUE #199: textDocument /documentSymbol is missing receiver functions
_From @josh-tepper on August 11, 2017 17:39_

When I set `"go.useLanguageServer": true`, I no longer see receiver functions (aka methods) when I type "@" in the command palette (though I see all other symbols).  This is in contrast to the behavior when I use go.docsTool set to either godoc or gogetdoc.  Nonetheless, I wish that I had these symbols because they're useful!

_Copied from original issue: Microsoft/vscode-go#1140_

keegancsmith:
Fixes #199

### DIFF
--- a/langserver/langserver_test.go
+++ b/langserver/langserver_test.go
@@ -928,6 +928,25 @@ type Header struct {
 				},
 			},
 		},
+		"recv in different file": {
+			rootURI: "file:///src/test/pkg",
+			fs: map[string]string{
+				"abc.go": `package a
+
+type XYZ struct {}
+`,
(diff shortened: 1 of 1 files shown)


keegancsmith:
To fix this, I think we should stop relying on `go/doc` and just directly parse the AST. The other option is to parse the whole package and then just filter out the symbols for the current file. However, I think that can have extra annoying corner cases we will have to deal with (not sure though).

PR #212, opened 2017-11-27
fix how end is calculated in textDocument/didChange

Contextualist:
The `Range.end` in LSP is exclusive. Namely, a `Range` struct specifies a range of characters from `Range.start` to the character just before `Range.end`.
Prior the fix, for example, an edit with the same start and end position will consume one character, though it shall take away none.


keegancsmith:
Thanks!

PR #221, opened 2018-01-02
Add function snippets to code completion results

pbitty:
This adds function args as a snippet when autocompleting function calls.

When a function call completion gets selected, the function args are included in the snippet to make it easier for a user to know what arguments need to be supplied.

Here's an example using Atom as the client:
func-snippet

(The above is using a local dev build of [ide-go](https://github.com/ckaznocha/ide-go) with completion support added.  I will be submitting a PR to add this support to ide-go.)

### DIFF
--- a/langserver/completion.go
+++ b/langserver/completion.go
@@ -62,20 +62,23 @@ func (h *LangHandler) handleTextDocumentCompletion(ctx context.Context, conn jso
 		case "var":
 			kind = lsp.CIKVariable
 		}
-		item := lsp.CompletionItem{
-			Label:  it.Name,
-			Kind:   kind,
-			Detail: it.Type,
+
+		itf, newText := h.getNewText(kind, it.Name, it.Type)
+		citems[i] = lsp.CompletionItem{
+			Label:            it.Name,
+			Kind:             kind,
+			Detail:           it.Type,
+			InsertTextFormat: itf,
+			// InsertText is deprecated in favour of TextEdit, but added here for legacy client support
+			InsertText: newText,
 			TextEdit: &lsp.TextEdit{
 				Range: lsp.Range{
 					Start: lsp.Position{Line: params.Position.Line, Character: params.Position.Character - rangelen},
 					End:   lsp.Position{Line: params.Position.Line, Character: params.Position.Character},
 				},
-				NewText: it.Name,
+				NewText: newText,
 			},
 		}
-		item = h.addInsertText(item)
-		citems[i] = item
 	}
 	return &lsp.CompletionList{
 		IsIncomplete: false,
@@ -83,19 +86,13 @@ func (h *LangHandler) handleTextDocumentCompletion(ctx context.Context, conn jso
 	}, nil
 }
 
-func (h *LangHandler) addInsertText(item lsp.CompletionItem) lsp.CompletionItem {
-    item.InsertTextFormat = lsp.ITFPlainText
-	if item.Kind == lsp.CIKFunction {
-		args := parseFuncArgs(item.Detail)
-        suffix := ""
-		if h.init.Capabilities.TextDocument.Completion.CompletionItem.SnippetSupport {
-			item.InsertTextFormat = lsp.ITFSnippet
-			args = genSnippetArgs(args)
-            suffix = "$0"
-		}
-		item.InsertText = fmt.Sprintf("%s(%s)%s", item.Label, strings.Join(args, ", "), suffix)
+func (h *LangHandler) getNewText(kind lsp.CompletionItemKind, name, detail string) (lsp.InsertTextFormat, string) {
+	if kind == lsp.CIKFunction && h.init.Capabilities.TextDocument.Completion.CompletionItem.SnippetSupport {
+		args := genSnippetArgs(parseFuncArgs(detail))
+		text := fmt.Sprintf("%s(%s)$0", name, strings.Join(args, ", "))
+		return lsp.ITFSnippet, text
 	}
-	return item
+	return lsp.ITFPlainText, name
 }
 
 func parseFuncArgs(def string) []string {
@@ -108,11 +105,11 @@ func parseFuncArgs(def string) []string {
 }
 
 func genSnippetArgs(args []string) []string {
-    newArgs := make([]string, len(args))
-    for i, a := range args {
-        // Closing curly braces must be escaped
-        a = strings.Replace(a, "}", "\\}", -1)
-        newArgs[i] = fmt.Sprintf("${%d:%s}", i+1, a)
-    }
-    return newArgs
+	newArgs := make([]string, len(args))
+	for i, a := range args {
+		// Closing curly braces must be escaped
+		a = strings.Replace(a, "}", "\\}", -1)
+		newArgs[i] = fmt.Sprintf("${%d:%s}", i+1, a)
+	}
+	return newArgs
 }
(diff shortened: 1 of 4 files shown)


Contextualist:
Good patch! 👍 I agree that some sort of hints on the arguments will help a lot. Just a question, how is this compared to `textDocument/signatureHelp` in LSP?

keegancsmith:
This is great! I am in two minds about this. Personally I'd prefer this to the signature helper, but it is treading on signature helpers feet and I could assume some users would not enjoy this.

However, auto complete support is very newly added. So maybe we can take some liberties and experiment with this and adjust the behaviour if users complains. What do you think @Contextualist?

pbitty:
Interesting!  I was not aware of signatureHelp.  It does seem there is some overlap between the features.

At a glance, it seems that in the above approach, go-langserver automatically supplies (forces?) the snippet on the client, whereas `signatureHelp` must be requested by the client and it's up to the client to decide to provide a snippet.

Conceptually, it sounds like letting the client decide would be more flexible in the long run, but the snippet approach would give this to clients for free, without requiring clients to implement the snippet behavior.

If you do decide to merge this, any thoughts on putting a flag around this feature, defaulted to enabled?

langserver/completion.go
```diff
@@ -71,9 +74,45 @@ func (h *LangHandler) handleTextDocumentCompletion(ctx context.Context, conn jso
...
+    item.InsertTextFormat = lsp.ITFPlainText
+	if item.Kind == lsp.CIKFunction {
+		args := parseFuncArgs(item.Detail)
+        suffix := ""
+		if h.init.Capabilities.TextDocument.Completion.CompletionItem.SnippetSupport {
+			item.InsertTextFormat = lsp.ITFSnippet
+			args = genSnippetArgs(args)
+            suffix = "$0"
+		}
+		item.InsertText = fmt.Sprintf("%s(%s)%s", item.Label, strings.Join(args, ", "), suffix)
```
Contextualist: `CompletionItem.InsertText` is deprecated. Simply override the value of `CompletionItem.TextEdit[0].NewText` here, which also applies the format specified in `CompletionItem.InsertTextFormat`.

Contextualist:
@keegancsmith, I would like to see this feature getting merged. Although personally I prefer signature help 😜, I can see many excited with this. And yes, if we decide to add a server-side flag, default to enabled.

Per option/flag issue, @pbitty, I am not sure whether it is the responsibility of server or that of client. For server, the option should be pushing function snippet or not; for client, applying function snippet or not (if not, no worry about the signature help because the user never gets a chance to type the triggering character `(`&`,`) Either way seems to be easy to implement.

Server-side flag seems "cleaner" to me (except for making the command line longer), plus I am not sure if the clients with snippet support has the option to disable function snippet (in long run, there might be other snippet supports (struct, func template, etc.)) or simply all snippet. Still, it may not be a huge issue, so it is up to you.

langserver/completion.go
Contextualist: Please do not modify the completion text if `SnippetSupport` is false. It would be annoying to delete the arguments snippet hint manually.

pbitty: Agreed.  I should have these changes done later today.

keegancsmith:
Lets get one more opinion before merging this in (and help decide if the flag should default to true or false). @ramya-rao-a what do you think of this feature? If we enable it by default everyone who uses the langserver in vscode-go may get a surprise.

lloiser:
FYI: The [go-plus](https://atom.io/packages/go-plus) package of atom has a setting to customise the completion suggestions:

* name: only the arg name
* name and type: show both arg name and it's type, like it is described in this PR (also my favourite 😉 )
* none: like in vscode

pbitty:
Sorry, I haven't had a chance to address the PR comments yet, but should be able to do it this weekend.  I will add a flag either way, and we can tweak its behavior pending the feedback on this PR.

main.go
```diff
@@ -66,6 +67,8 @@ func main() {
 
 	langserver.GocodeCompletionEnabled = *gocodecompletion
 
+	langserver.FuncSnippetEnabled = *funcSnippetEnabled
```
pbitty: Not directly related to this PR, but I feel like it would help keep the package code more isolated if the above flags were put into `Config` object and passed into `langserver.NewHandler()`.  Is there any interest in this?

langserver/completion.go
```diff
@@ -59,16 +63,21 @@ func (h *LangHandler) handleTextDocumentCompletion(ctx context.Context, conn jso
...
 		citems[i] = lsp.CompletionItem{
-			Label:  it.Name,
-			Kind:   kind,
-			Detail: it.Type,
+			Label:            it.Name,
+			Kind:             kind,
+			Detail:           it.Type,
+			InsertTextFormat: itf,
+			// InsertText is deprecated in favour of TextEdit, but added here for legacy client support
+			InsertText: newText,
```
pbitty: Would you want to keep `InsertText` around for older clients?

main.go
```diff
@@ -20,16 +20,17 @@ import (
...
+	addr               = flag.String("addr", ":4389", "server listen address (tcp)")
+	trace              = flag.Bool("trace", false, "print all requests and responses")
+	logfile            = flag.String("logfile", "", "also log to this file (in addition to stderr)")
+	printVersion       = flag.Bool("version", false, "print version and exit")
+	pprof              = flag.String("pprof", ":6060", "start a pprof http server (https://golang.org/pkg/net/http/pprof/)")
+	freeosmemory       = flag.Bool("freeosmemory", true, "aggressively free memory back to the OS")
+	usebinarypkgcache  = flag.Bool("usebinarypkgcache", true, "use $GOPATH/pkg binary .a files (improves performance)")
+	maxparallelism     = flag.Int("maxparallelism", -1, "use at max N parallel goroutines to fulfill requests")
+	gocodecompletion   = flag.Bool("gocodecompletion", false, "enable completion (extra memory burden)")
+	funcSnippetEnabled = flag.Bool("func-snippet-enabled", true, "enable argument snippets on func completion")
```
pbitty: Defaulted to `true` for now, until this is decided.

langserver/completion.go
pbitty: My vote is to keep this around for the time being.  The Atom client does not yet read snippet content from `CompletionItem.textEdit`.  I am not sure about other clients.

I've opened a [PR](https://github.com/atom/atom-languageclient/pull/165) to fix it.

langserver/completion.go
```diff
@@ -12,6 +14,8 @@ import (
 var (
 	GocodeCompletionEnabled = false
 	CIKConstantSupported    = lsp.CIKVariable // or lsp.CIKConstant if client supported
+	FuncSnippetEnabled      = false
+	funcArgsRegexp          = regexp.MustCompile("func\\(([^)]+)\\)")
```
keegancsmith: nit: you can use backquote string to make this regex more readable

```go
`func\(([^)]+)\)`
```

main.go
keegancsmith: given gocodecompletion defaults to false, this sounds good to me.

main.go
keegancsmith: yes, the organic growth of how this package works has led to this uglyness. Cleaning up sounds great to me.

ramya-rao-a:
Sorry folks, it took me a while to respond here.

The Go extension to VS Code has similar settings as mentioned in https://github.com/sourcegraph/go-langserver/pull/221#issuecomment-355532160. If we can support the 3 modes, then that it would be great

Issue #223, opened 2018-01-08
Hover fails for methods on types


ramya-rao-a:
Hovering on `b.Bye` in the below code in VS Code fails

```
package main

import (
	"fmt"
)

func main() {

	b := &Hello{
		a: 1,
	}

	fmt.Println(b.Bye())
}

type Hello struct {
	a int
}

func (h *Hello) Bye() int {
	return h.a
}
```

<img width="930" alt="screen shot 2018-01-07 at 5 53 16 pm" src="https://user-images.githubusercontent.com/16890566/34656789-addaa26a-f3d3-11e7-9959-d88cce0e8ce9.png">

keegancsmith:
Thanks, this looks easy to fix. Will add to the test suite.

PR #224, opened 2018-01-09
Include docs on methods for godef hover

keegancsmith:
Fixes https://github.com/sourcegraph/go-langserver/issues/223


keegancsmith:
Given this has tests and I know @slimsag is quite busy at the moment, I'm gonna merge in now and rely on post-merge review.

PR #226, opened 2018-01-11
Add config struct

pbitty:
Following up on https://github.com/sourcegraph/go-langserver/pull/221#discussion_r160032897, this PR adds an explicit `Config` struct for grouping `langserver` config settings, and removes global config vars.

### DIFF
--- a/langserver/config.go
+++ b/langserver/config.go
@@ -12,4 +12,9 @@ var (
 )
 
 type Config struct {
+	// MaxParallelism controls the maximum number of goroutines that should be used
+	// to fulfill requests. This is useful in editor environments where users do
+	// not want results ASAP, but rather just semi quickly without eating all of
+	// their CPU.
+	MaxParallelism int
 }
--- a/langserver/config.go
+++ b/langserver/config.go
@@ -23,4 +23,7 @@ type Config struct {
 	// not want results ASAP, but rather just semi quickly without eating all of
 	// their CPU.
 	MaxParallelism int
+	// UseBinaryPkgCache controls whether or not $GOPATH/pkg binary .a files should
+	// be used.
+	UseBinaryPkgCache bool
 }
--- a/langserver/config.go
+++ b/langserver/config.go
@@ -12,6 +12,8 @@ var (
 )
 
 type Config struct {
+	// GocodeCompletionEnabled enables code completion feature (using gocode)
+	GocodeCompletionEnabled bool
 	// MaxParallelism controls the maximum number of goroutines that should be used
 	// to fulfill requests. This is useful in editor environments where users do
 	// not want results ASAP, but rather just semi quickly without eating all of
--- a/langserver/config.go
+++ b/langserver/config.go
@@ -10,3 +10,6 @@ var (
 	// worth it.
 	envWarmupOnInitialize = os.Getenv("GOLSP_WARMUP_ON_INITIALIZE")
 )
+
+type Config struct {
+}
--- a/langserver/config.go
+++ b/langserver/config.go
@@ -12,6 +12,10 @@ var (
 )
 
 type Config struct {
+	// FuncSnippetEnabled enables the returning of enable argument snippets
+	// on `func` completions, eg. func(foo string, arg2 bar).
+	// Requires code completion to be enabled.
+	FuncSnippetEnabled bool
 	// GocodeCompletionEnabled enables code completion feature (using gocode)
 	GocodeCompletionEnabled bool
 	// MaxParallelism controls the maximum number of goroutines that should be used
--- a/langserver/config.go
+++ b/langserver/config.go
@@ -27,3 +27,9 @@ type Config struct {
 	// be used.
 	UseBinaryPkgCache bool
 }
+
+func NewDefaultConfig() Config {
+	return Config{
+		MaxParallelism: 8,
+	}
+}
--- a/langserver/langserver_test.go
+++ b/langserver/langserver_test.go
@@ -1046,7 +1046,10 @@ func TestServer(t *testing.T) {
 			}
 
 			h := &LangHandler{
-				config:        Config{GocodeCompletionEnabled: true},
+				config: Config{
+					FuncSnippetEnabled:      true,
+					GocodeCompletionEnabled: true,
+				},
 				HandlerShared: &HandlerShared{},
 			}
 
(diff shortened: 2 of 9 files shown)


langserver/langserver_test.go
```diff
@@ -1215,16 +1219,16 @@ func lspTests(t testing.TB, ctx context.Context, fs *AtomicFS, c *jsonrpc2.Conn,
 		wantGodefHover = cases.wantHover
 	}
 
-	if len(wantGodefDefinition) > 0 || (len(wantGodefHover) > 0 && fs != nil) || len(cases.wantCompletion) > 0 {
-		UseBinaryPkgCache = true
+	if len(wantGodefDefinition) > 0 || (len(wantGodefHover) > 0 && h != nil) || len(cases.wantCompletion) > 0 {
+		h.config.UseBinaryPkgCache = true
```
pbitty: There may be a good case for refactoring this more, to avoid mutating the config mid-test, but I felt this was out of scope of this PR.

langserver/config.go
```diff
@@ -10,3 +10,26 @@ var (
...
+	// on `func` completions, eg. func(foo string, arg2 bar).
+	// Requires code completion to be enabled.
+	FuncSnippetEnabled bool
+	// GocodeCompletionEnabled enables code completion feature (using gocode)
+	GocodeCompletionEnabled bool
+	// MaxParallelism controls the maximum number of goroutines that should be used
+	// to fulfill requests. This is useful in editor environments where users do
+	// not want results ASAP, but rather just semi quickly without eating all of
+	// their CPU.
+	MaxParallelism int
```
pbitty: `MaxParallelism` doesn't seem to be used anywhere.  Should it be removed?

keegancsmith: Nice catch. Yes it can be removed.

langserver/langserver_test.go
```diff
@@ -1216,16 +1219,16 @@ func lspTests(t testing.TB, ctx context.Context, fs *AtomicFS, c *jsonrpc2.Conn,
 		wantGodefHover = cases.wantHover
 	}
 
-	if len(wantGodefDefinition) > 0 || (len(wantGodefHover) > 0 && fs != nil) || len(cases.wantCompletion) > 0 {
-		UseBinaryPkgCache = true
+	if len(wantGodefDefinition) > 0 || (len(wantGodefHover) > 0 && h != nil) || len(cases.wantCompletion) > 0 {
```
keegancsmith: shouldn't you be checking that `h.FS != nil`

langserver/langserver_test.go
keegancsmith: Given this was mutating a global variable and never returning it to the old value, this is actually better. Maybe we can just always have `UseBinaryPkgCache` set to true?

langserver/config.go
```diff
@@ -27,3 +27,9 @@ type Config struct {
 	// be used.
 	UseBinaryPkgCache bool
 }
+
+func NewDefaultConfig() Config {
+	return Config{
+		MaxParallelism: 8,
+	}
```
keegancsmith: given we can remove MaxParallelism, rather get rid of new default config then we can just rely on the zero value for Config.

langserver/langserver_test.go
```diff
@@ -1038,16 +1038,21 @@ func (h *Hello) Bye() int {
...
 	for label, test := range serverTestCases {
 		t.Run(label, func(t *testing.T) {
 			if test.skip {
 				t.Skip()
 				return
 			}
 
-			h := &LangHandler{HandlerShared: &HandlerShared{}}
+			cfg := NewDefaultConfig()
+			cfg.FuncSnippetEnabled = true
```
lloiser: This wasn't `true` previously. But the build is green so I assume it does no harm right?

langserver/config.go
pbitty: Yup, will do.

langserver/langserver_test.go
pbitty: It gets reverted to `false` at [langserver_test.go:1277](https://github.com/pbitty/go-langserver/blob/68714b56f864bd53c7e80faa132e9d30c0c173e2/langserver/langserver_test.go#L1277).

langserver/langserver_test.go
pbitty: Prior to this PR, it seems `fs` being `nil` was a signal to not perform the tests in the `if` block.  See [integration_test.go](https://github.com/pbitty/go-langserver/blob/add-config-struct/langserver/integration_test.go#L76-L137).

In this case, we're passing the `LangHandler` instead of `FS`, but approach remains the same.  The way I see it, I sorta hacked around the issue and it is now a little bit harder to read.

I'm happy to submit another PR to refactor these specific tests.  Maybe we can break out this block into its own func with a clear name, and only call it where it makes sense?

langserver/config.go
Contextualist: Well... It should have been used.

https://github.com/sourcegraph/go-langserver/blob/a28dd2535cb72e3748c9b0292565f26a25046be9/langserver/symbol.go#L343-L351

Notice the magic number `8` in L351. Look like it might originally meant to be `MaxParallelism`. But I am not sure if such customization is still needed.

langserver/langserver_test.go
pbitty: Good catch.  Since `FuncSnippetEnabled` is defaulted to `true` in `main.go`, I wonder if we should keep it `true` here also, since `GocodeCompletionEnabled` is `true` in this test.

It might make little difference in practice, but it does make the two worlds match a little more.

Any preference or another suggestion?

langserver/config.go
pbitty: Let me know which way you'd like to go and I'll keep or remove the config var.

langserver/langserver_test.go
keegancsmith: aah I missed that. This this seems good to me.

langserver/config.go
keegancsmith: hah. Ok lets use it at this point.

langserver/langserver_test.go
keegancsmith: I'm happy to accept any PR you send which cleans stuff up :) Checking `h != nil` LGTM

langserver/langserver_test.go
keegancsmith: Yeah leave it true.

pbitty:
Updated it to use`MaxParallelism`.

Issue #229, opened 2018-01-23
Result is null when find references


Cosson2017:
Result is null when find references.
follow is log:
![screenshot_20180124_112622](https://user-images.githubusercontent.com/20379044/35313232-3411fa62-00fa-11e8-98df-2f0c23ee923c.png)

keegancsmith:
Thanks, we have that issue a few times due to how go json serializes empty values. I'll take a look for the reference code.

PR #230, opened 2018-01-31
fix references result is empty

Cosson2017:
fix references result is empty


Cosson2017:
#229

keegancsmith:
@Cosson2017 Sorry I missed this! I pushed an alternative fix. Does the latest master solve your problem?

Cosson2017:
deepinscreenshot_select-area_20180222144657
I find h.init.RootImportPath is "" causes function pkgInWorkspace return false. And when I let pkgInWorkspace return true, reference list responsed is not empty.

ramya-rao-a:
@keegancsmith I can attest to that. It definitely did work sometime last year when we added support to the language server in VS Code

PR #233, opened 2018-02-09
wire connection through

slimsag:
The goal of this PR is to pass the JSONRPC2 connection through to `langserver.FindPackageFunc` which allows implementations of `FindPackageFunc` to use the JSONRPC2 connection to access the `xcache` extension.

### DIFF
--- a/langserver/handler_shared.go
+++ b/langserver/handler_shared.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/sourcegraph/ctxvfs"
+	"github.com/sourcegraph/jsonrpc2"
 )
 
 // HandlerShared contains data structures that a build server and its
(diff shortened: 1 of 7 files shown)


keegancsmith:
The build failure is unrelated. Seems to be a break a week ago on windows. cc @sqs

langserver/handler_shared.go
```diff
@@ -25,9 +26,9 @@ type HandlerShared struct {
 
 // FindPackageFunc matches the signature of loader.Config.FindPackage, except
 // also takes a context.Context.
-type FindPackageFunc func(ctx context.Context, bctx *build.Context, importPath, fromDir string, mode build.ImportMode) (*build.Package, error)
+type FindPackageFunc func(ctx context.Context, bctx *build.Context, conn jsonrpc2.JSONRPC2, importPath, fromDir string, mode build.ImportMode) (*build.Package, error)
```
keegancsmith: Can you update the doc string to mention conn. It would also be nice if you made it the 2nd param, such that everything after the 2nd param matched FindPackage.

slimsag:
Yep, your solution to this issue was much better. 🎉

PR #235, opened 2018-02-13
Vendor dependencies

keegancsmith:
We are mainly consumed as a command, so we should prevent future breakages and
vendor our changes in. The main consumer of us as a library is Sourcegraph, but
dep should flatten the deps in that case. We also declare no constraints in
Gopkg.toml to prevent it influencing anyone who uses go-langserver as a
library.


keegancsmith:
hmm, this seems to break CI. @slimsag If you think this is a good idea, I'll fix up the tests.

keegancsmith:
@slimsag I have no idea how to fix the umatching-imports failing. I think its because the test previously assumed log15 would be on the GOPATH, but having it in vendor stopped that from being installed? I tried adjusting the import path with no luck. I have just removed the test. If you are fine with that I'll merge in.

slimsag:
I'll see if I can fix it today. I think that's the only test we have that ensures e.g. referencing `log15` works when the imported package path is `log15.v2` / doesn't match. IIRC we had some bugs with this in the past, so I think it's an important test to keep.

slimsag:
The fix was a bit counterintuitive, but better than removing the test :)

PR #236, opened 2018-02-13
Test tip on travis

keegancsmith:
We need to sort out our tests, they currently fail.

### DIFF
--- a/langserver/langserver_test.go
+++ b/langserver/langserver_test.go
@@ -431,7 +431,7 @@ package main; import "test/pkg"; func B() { p.A(); B() }`,
 	"goroot": {
 		rootURI: "file:///src/test/pkg",
 		fs: map[string]string{
-			"a.go": `package p; import "fmt"; var _ = fmt.Println; var x int`,
+			"a.go": `package p; import "fmt"; var _ = fmt.Println; var x int; type _ = fmt.State`,
 		},
 		mountFS: map[string]map[string]string{
 			"/goroot": {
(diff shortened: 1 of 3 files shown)


keegancsmith:
@Contextualist do you mind taking a look? The only thing failing now on go1.10 is completiion:

```
--- FAIL: TestServer (11.77s)
    --- FAIL: TestServer/go_vendored_dep (0.61s)
        langserver_test.go:1265: $ go install -v ...
                test/pkg/vendor/github.com/v/vendored
                test/pkg
        --- FAIL: TestServer/go_vendored_dep/completion-a.go:1:34 (0.00s)
                langserver_test.go:1422: got "", want "1:20-1:34 github.com/v/vendored module "
        --- FAIL: TestServer/go_vendored_dep/completion-a.go:1:61 (0.00s)
                langserver_test.go:1422: got "", want "1:61-1:61 V function func()"
```

langserver/langserver_test.go
```diff
@@ -449,7 +449,7 @@ package main; import "test/pkg"; func B() { p.A(); B() }`,
 				// "a.go:1:53": "type int int",
 			},
 			overrideGodefDefinition: map[string]string{
-				"a.go:1:40": "/goroot/src/fmt/print.go:256:6-256:13",  // hitting the real GOROOT
+				"a.go:1:71": "/goroot/src/fmt/print.go:38:6-38:11",    // hitting the real GOROOT
```
lloiser: Every small change in the src will fail this test case. Is there a better approach to this? Maybe just assert that a _valid_ file and line number is returned?

keegancsmith: yeah that sounds much better. It requires a bit of cleaning up in our tests, but that seems fine given this is relying on GOROOT of the user running the tests. Right now the location I have picked hasn't changed in 2 years.

keegancsmith: Updated to test path instead for this specific test.

Contextualist:
@keegancsmith Ah! Thanks for fixing the vendor issue. I just found that .a files for vendor packages are missing, never had thought that it is the problem due to go 1.10 package caching.

keegancsmith:
@Contextualist yeah sorry I should of mentioned I decided to look into it.

Contextualist:
@keegancsmith That's totally fine. 😄 Appreciate your efficiency!

Nonetheless, we have to wait for `gocode` to fix the issue, otherwise go1.10 users have to either install vendor package manually or disable package caching.

Issue #216, opened 2017-12-08
Change deprecated MarkedString to MarkdownString


BugDiver:
Language server protocol has deprecated the `MarkedString` and they are replacing it with `MarkdownString`.
Please update the lsp library.
Ref  https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#new-markupcontent

BugDiver:
@keegancsmith I have created a [PR](#237) for it. Please have a look and provide your valuable feedback.

PR #238, opened 2018-02-15
Added MarkdownString type, Deprecated MarkedString #216

BugDiver:
Added interface MarkupContent for backword compatibility

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -59,7 +59,7 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 		r := rangeForNode(fset, node)
 		if pkgName := packageStatementName(fset, pkg.Files, node); pkgName != "" {
 			return &lsp.Hover{
-				Contents: maybeAddComments(comments, []lsp.MarkedString{{Language: "go", Value: "package " + pkgName}}),
+				Contents: maybeAddComments(comments, []lsp.MarkupContent{lsp.MarkedString{Language: "go", Value: "package " + pkgName}}),
 				Range:    &r,
 			}, nil
 		}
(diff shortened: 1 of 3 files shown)


BugDiver:
Can someone please check why the build is broken. It's passing for `appveyor`. Two out of three builds are passing for `travis` as well. Maybe retriggering fixes that.

Any other comments/feedback?

keegancsmith:
It triggered a data race which I haven't seen before! I filed https://github.com/sourcegraph/go-langserver/issues/244 but it looks very much unrelated to your code. Restarted the CI job. Thanks for the ping.

keegancsmith:
Thanks, I just read the spec for `MarkupContent`. It seems a hover can now return

```typescript
contents: MarkedString | MarkedString[] | MarkupContent
```

and that `MarkupContent` is somewhat equivalent to 
```go
type MarkupContent struct {
  Kind string // 'plaintext' | 'markdown'
  Value string
}
```

I'm almost tempted to drop support for the deprecated `MarkedString` to simplify. But that would likely be painful in practice. So we should adjust this PR so there is a clear path forward for deprecating. There isn't a nice way to convert `MarkedString[]` into `MarkupContent` while still remaining correct, so I think a type with an awful name like `MarkedStringOrMarkupContent` is unfortunately the way forward, then we can fully remove `MarkedString` support at some point.

Also your PR supports raw strings being converted into `MarkupContent` which isn't needed. It also has a field called `IsTrusted` which I don't see in the spec.

robbyoconnor:
rebase.

ianlopshire:
What is preventing `MarkedString[]` from being converted to `MarkupContent`?

The `MarkedString` spec doesn't differentiate between `markdown` and `plaintext` so we can just default to `markdown` when converting.



```go
// untested example
func MarkedStringsToMarkupContent(ms []MarkedString) (MarkupContent) {
	var b strings.Builder
	for i := range ms {
		if ms[i].isRawString {
			b.WriteString(ms[i].Value)
			continue
		}
		b.WriteString("```")
		b.WriteString(ms[i].Language)
                b.WriteString("\n")
		b.WriteString(ms[i].Value)
                b.WriteString("\n")
		b.WriteString("```")
		
		if i != len(ms) -1 {
			b.WriteString("\n")
		}
	}

	return MarkupContent{
		Kind: "markdown",
		Value: b.String()
	}
}
```

PR #242, opened 2018-02-24
differentiate between interface and class symbols

### LINKED ISSUE #241: No difference between interfaces and classes
Hi, this probably ends up being more of a feature request than anything else. But I noticed adding an interface type to the [test here](https://github.com/sourcegraph/go-langserver/blob/1e571d690e259cdb944b96c5d3596dff9e6cb13d/langserver/langserver_test.go#L779) (such as `type UVW interface{}`) returns it as type class (specifically `"/src/test/pkg/abc.go:class:UVW:3:6"`). 

Since the lsp has a value for [interfaces](https://github.com/sourcegraph/go-langserver/blob/1e571d690e259cdb944b96c5d3596dff9e6cb13d/pkg/lsp/service.go#L401) I'd expect this to return an interface type. Specifically, this seems to lump structs and interfaces into the same category on downstream clients (e.g. Symbol lookup in vscode).

vitreuz:
Fixes #241

### DIFF
--- a/langserver/symbol.go
+++ b/langserver/symbol.go
@@ -444,7 +444,7 @@ func astPkgToSymbols(fs *token.FileSet, astPkg *ast.Package, buildPkg *build.Pac
 	// Emit decls
 	var pkgSyms []symbolPair
 	for _, t := range docPkg.Types {
-		pkgSyms = append(pkgSyms, toSym(t.Name, buildPkg, "", lsp.SKClass, fs, declNamePos(t.Decl, t.Name)))
+		pkgSyms = append(pkgSyms, toSym(t.Name, buildPkg, "", typeSpecSym(t), fs, declNamePos(t.Decl, t.Name)))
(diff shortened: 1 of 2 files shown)


langserver/symbol.go
```diff
@@ -479,6 +479,23 @@ func astPkgToSymbols(fs *token.FileSet, astPkg *ast.Package, buildPkg *build.Pac
 	return pkgSyms
 }
 
+func typeSpecSym(t *doc.Type) lsp.SymbolKind {
+	// This usually has one, but we're running through a for loop in case it has
+	// none. In either case, the default is an SKClass.
+
+	// NOTE: coincidentally, doing this gives access to the methods for an
+	// interface and fields for a struct. Possible solution to Github issue #36?
```
keegancsmith: ooh, nice. It does indeed sound like it would do that.

keegancsmith:
@vitreuz Thanks so much for the contribution!

PR #243, opened 2018-02-26
langserver/internal/tools: make this package externally accessible

sqs:
```
gomvpkg -from github.com/sourcegraph/go-langserver/langserver/internal/tools -to github.com/sourcegraph/go-langserver/pkg/tools
```

I would like to use this package from another repository.


sqs:
Renamed to `tools` from `toolsutils`.

PR #246, opened 2018-03-16
Optimize global pkglevel references

josharian:
This is a lot of new code.

I've tried to document it carefully, but I will invariably have failed somewhere. Questions are very welcome, including naive questions; I will update the docs in the code with answers.

I originally wrote this change in the x/tools repo, as https://golang.org/cl/97800. Porting it to go-langserver exposed so many bugs that I've done this next round here instead, with some pretty significant changes. (What lovely tests you have!) A vigorous and skeptical review is welcome.

Once this gets a thumbs up, I will backport this new and improved version to x/tools for @alandonovan to review. If you'd prefer, we can wait to actually merge this until Alan is happy with the x/tools version as well.

There is opportunity for further straightforward optimization here by parallelizing some of this work. Given the considerable subtlety involved in getting this right in the first place, I will parallelize separately as follow-up work.

cc @keegancsmith for review

cc @sqs as fyi

### DIFF
--- a/langserver/references.go
+++ b/langserver/references.go
@@ -1,6 +1,7 @@
 package langserver
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -10,11 +11,14 @@ import (
 	"go/parser"
 	"go/token"
 	"go/types"
+	"io"
 	"math"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
 
+	"golang.org/x/tools/go/buildutil"
 	"golang.org/x/tools/go/loader"
 	"golang.org/x/tools/refactor/importgraph"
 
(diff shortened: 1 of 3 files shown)


langserver/references.go
```diff
@@ -447,6 +465,241 @@ func findReferences(ctx context.Context, lconf loader.Config, pkgInWorkspace fun
...
+		uIsXTest := strings.HasSuffix(u, "!test") // indicates whether this package is the special defpkg xtest package
+		u = strings.TrimSuffix(u, "!test")
+
+		if !pkgInWorkspace(u) {
+			continue
+		}
+
+		// Resolve package.
+		// TODO: is fromDir == "" correct?
+		pkg, err := find(ctx, bctx, u, "", build.IgnoreVendor)
```
keegancsmith: `fromDir` can be empty if we don't have relative import paths (which I don't think we should since users is from the importgraph?) https://sourcegraph.com/github.com/golang/go@go1.9/-/blob/src/go/build/build.go#L527

keegancsmith:
Thank you for the detailed comments btw, it made it really easy to follow along.

PR #83, opened 2017-01-12
WIP langserver: TextEdits computed from diff for gofmt

keegancsmith:
Before this commit we would send an edit which deleted everything followed up by
inserting the formatted code. This lead to a bad experience in vscode. This
change implemented a more minimal number of TextEdits by using difflib. The
behaviour now matches the behaviour of vscode-go.

WIP Known Issues
- [ ] Trailing whitespace is not completely removed
- [ ] test harness needs updating
- [ ] more testing in vscode
- [ ] order of textedits may be incorrect

Fixes #76

PR #248, opened 2018-03-17
gofmt improvements

lloiser:
This PR enhances #83 

It attempts to fix the following problems:
* URI handling in windows was broken due to inconsistent use of lower-case and not lower-case paths throughout the server
* correctly apply changes of `textDocument/didChange` to the cached file
* small changes to the formatting code to be better testable

All of these changes are backed by tests, so 🤞 that I didn't break anything 😁


langserver/fs.go
```diff
@@ -124,14 +124,24 @@ func (h *overlay) didChange(params *lsp.DidChangeTextDocumentParams) error {
...
+	return nil
+}
+
+// ApplyContentChanges updates `contents` based on `changes`
+func ApplyContentChanges(uri lsp.DocumentURI, contents []byte, changes []lsp.TextDocumentContentChangeEvent) ([]byte, error) {
+	for _, change := range changes {
 		if change.Range == nil && change.RangeLength == 0 {
-			contents = []byte(change.Text) // new full content
-			continue
+			return []byte(change.Text), nil // new full content
```
keegancsmith: why do you return here, there may be more changes?

langserver/fs.go
```diff
@@ -140,11 +150,11 @@ func (h *overlay) didChange(params *lsp.DidChangeTextDocumentParams) error {
 			// RangeLength not specified, work it out from Range.End
 			end, ok, why = offsetForPosition(contents, change.Range.End)
 			if !ok {
-				return fmt.Errorf("received textDocument/didChange for invalid position %q on %q: %s", change.Range.Start, params.TextDocument.URI, why)
+				return nil, fmt.Errorf("received textDocument/didChange for invalid position %q on %q: %s", change.Range.Start, uri, why)
 			}
 		}
-		if start < 0 || end >= len(contents) || end < start {
-			return fmt.Errorf("received textDocument/didChange for out of range position %q on %q", change.Range, params.TextDocument.URI)
+		if start < 0 || end > len(contents) || end < start {
```
keegancsmith: nice catch.

langserver/fs.go
```diff
@@ -124,14 +124,24 @@ func (h *overlay) didChange(params *lsp.DidChangeTextDocumentParams) error {
...
+	if err != nil {
+		return err
+	}
+
+	h.set(params.TextDocument.URI, contents)
+	return nil
+}
+
+// ApplyContentChanges updates `contents` based on `changes`
+func ApplyContentChanges(uri lsp.DocumentURI, contents []byte, changes []lsp.TextDocumentContentChangeEvent) ([]byte, error) {
```
keegancsmith: Can you not export this function? I don't think we have a need to use it outside of this pkg yet.

langserver/langserver_test.go
```diff
@@ -30,6 +30,7 @@ import (
 
 type serverTestCase struct {
 	skip    bool
+	focus   bool
```
keegancsmith: go supports this already since each test case is a subtest. You just need to do `go test -run TestServer/focusregex`

keegancsmith:
Also I haven't looked too deeply into it, but CI picked up a race condition inside of the integration test. Given you modified the integration test it's likely related to this PR (although may just be an old problem that is rare to trigger)

keegancsmith:
oops, just noticed you referenced the PR I mentioned in the review in your description.

langserver/fs_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/sourcegraph/go-langserver/pkg/lsp"
+)
+
+type applyContentChangesTestCase struct {
+	code     string
+	changes  []lsp.TextDocumentContentChangeEvent
+	expected string
+}
+
+var applyContentChangesTestCases = map[string]applyContentChangesTestCase{
```
Contextualist: Ah, after you extract the content change calculation, these tests look much better than my compromised one in 325afcc18412b620414884bc305f1748c70205ac. You may revert that if you please, because they basically aim to do the same thing.

langserver/langserver_test.go
lloiser: I basically just used it to focus the tests in code so that I can easily debug the tests using https://github.com/derekparker/delve (which btw works perfectly without any problems!)
I will revert it :)

lloiser:
I was able to pin down the race (reading the stacktraces, not on my machine...). It happens because of this:

https://github.com/sourcegraph/go-langserver/blob/8956eb2bc9b3c309c3bcb95c81a2a0d47fed7b13/langserver/handler.go#L396

Yet I'm a bit out of ideas how to fix this... I assume this requires a rewrite of this async type checking to some degree...

langserver/fs_test.go
lloiser: I think it's a good idea to keep it in because it asserts the behaviour on a higher level (integration test) in addition to my unit tests.

keegancsmith:
ok that race condition is unrelated to your changes. I actually thought I fixed that issue where we modified `build.Default` for integration tests. I'll see if I can find the change I had for that.

PR #250, opened 2018-03-26
support textDocument/implementation (Go interface impls queries)

sqs:
This implements the LSP textDocument/implementation method; see https://microsoft.github.io/language-server-protocol/specification#textDocument_implementation.

It adds the following functionality:

- on an interface, shows the types that implement it
- on a non-interface type, shows the interfaces it implements
- on an interface method, shows the methods (in implementing types) that implement it
- on a non-interface method, shows the interface methods it implements

Known problems:

- It copies a lot more code from guru.
- It could also be made more efficient in the future (it does some duplicate work right now).
- It does not let the user specify the analysis scope (as they can in guru). It uses the forward/reverse import graph.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,4 +1,4 @@
-Portions of code are adapted from go/types and github.com/golang/gddo:
+Portions of code are adapted from go/types, golang.org/x/tools, and github.com/golang/gddo:
 
 Copyright (c) 2013 The Go Authors. All rights reserved.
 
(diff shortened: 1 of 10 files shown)


keegancsmith:
@sqs build seems to be broken on windows with this PR. I don't know why, it seems like it is failing for an infrastructure reason, rather than your change.

sqs:
#252 passes, so use that PR instead (based on master, not no-godef branch)

PR #251, opened 2018-03-26
completion: gracefully handle unexpected request

Contextualist:
Panic occurs when a mistakenly implemented client requests for a completion despite the langserver had set `InitializeResult.capabilities.completionProvider` to false. (see onivim/oni#1378)

This change prohibits the langserver from processing the request with an uninitiated gocode daemon and offers helpful error message.


keegancsmith:
thanks! 🍻

PR #253, opened 2018-03-29
Publish diagnostics

lloiser:
This PR adds a very basic cache that holds diagnostics for each checked file. A file entry is removed once no more diagnostics exist for it.
An async `typecheck` now also happens on every save requests.

A simple unit test asserts the cache updating and an integration test modifies and saves files and then asserts certain diagnostics to be sent to the client.

Unfortunately it's currently based on #248 because it contains some changes that are necessary to work on my _windows_ machine 😞 

This is probably not 100% ready to be merged yet, but I think it is a good starting point 😁

### DIFF
--- a/langserver/diagnostics.go
+++ b/langserver/diagnostics.go
@@ -7,6 +7,7 @@ import (
 	"go/token"
 	"go/types"
 	"strings"
+	"sync"
 
 	"golang.org/x/tools/go/loader"
 
@@ -18,17 +19,36 @@ import (
 
 type diagnostics map[string][]*lsp.Diagnostic // map of URI to diagnostics (for PublishDiagnosticParams)
 
+type diagnosticsCache struct {
+	cache diagnostics
+	lock  sync.Mutex
+}
+
+func (p *diagnosticsCache) update(fn func(diagnostics) diagnostics) {
+	p.lock.Lock()
+	if p.cache == nil {
+		p.cache = diagnostics{}
+	}
+	p.cache = fn(p.cache)
+	p.lock.Unlock()
+}
+
+func newDiagnosticsCache() *diagnosticsCache {
+	return &diagnosticsCache{
+		cache: diagnostics{},
+	}
+}
+
 // publishDiagnostics sends diagnostic information (such as compile
 // errors) to the client.
-func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSONRPC2, diags diagnostics) error {
-	// Our diagnostics are currently disabled because they behave
-	// incorrectly. We do not keep track of which files have failed /
-	// succeeded, so we do not send empty diagnostics to clear compiler
-	// errors/etc. https://github.com/sourcegraph/go-langserver/issues/23
-	// Leaving the code here for when we do actually fix this.
-	if cake := false; !cake {
-		return nil
+func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSONRPC2, diags diagnostics, files []string) error {
+	if diags == nil {
+		diags = diagnostics{}
 	}
+
+	h.diagnosticsCache.update(func(cached diagnostics) diagnostics {
+		return updateCachedDiagnostics(cached, diags, files)
+	})
 
 	for filename, diags := range diags {
 		params := lsp.PublishDiagnosticsParams{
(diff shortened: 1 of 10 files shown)


lloiser:
sry for the delay. I'm aware of the build/test problems. Let's see if I can fix them in the next few days.

langserver/integration_test.go
```diff
@@ -217,11 +337,74 @@ func integrationTest(
 		}
 	}
 
-	ctx := context.Background()
+	cfg := NewDefaultConfig()
+	cfg.UseBinaryPkgCache = true
```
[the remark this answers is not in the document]
keegancsmith: Sorry I haven't looked at this yet! But cc @slimsag about UseBinaryPkgCache question.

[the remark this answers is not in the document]
slimsag: The idea with `UseBinaryPkgCache` was that you are already running `go install` on your own. This concept actually comes from https://github.com/rogpeppe/godef -- which makes the same assumption. If you are not running `go install` on your code, however, then this option will not work well for you (you'll either get no results or stale results). This feature exists primarily because `go/loader` does not allow for caching type-checking results.

In your case here, I think you should just not set `UseBinaryPkgCache = true`. That means it will not use the cache, and will instead use the other codepath that performs type-checking on-the-fly (which is slower, but ideal for a test such as this)

lloiser:
I'm not sure if this PR still makes sense if we assume that diagnostics are served by another editor plugin... What's your opinion on that?

keegancsmith:
We can have another editor plugin do it, but it does make sense for the langserver to properly support diagnostics.

langserver/diagnostics.go
```diff
@@ -18,18 +19,37 @@ import (
 
 type diagnostics map[string][]*lsp.Diagnostic // map of URI to diagnostics (for PublishDiagnosticParams)
 
+type diagnosticsCache struct {
+	cache diagnostics
+	lock  sync.Mutex
```
keegancsmith: nit: use style of a "mutex hat", so call rename `lock` to `mu` and have it above cache. https://dmitri.shuralyov.com/idiomatic-go#mutex-hat

langserver/diagnostics.go
```diff
@@ -18,18 +19,37 @@ import (
...
+}
+
+func (p *diagnosticsCache) update(fn func(diagnostics) diagnostics) {
```
keegancsmith: could you document this behaviour here? That we enforce that only one fn is running at a time.

langserver/handler.go
```diff
@@ -392,7 +397,7 @@ func (h *LangHandler) Handle(ctx context.Context, conn jsonrpc2.JSONRPC2, req *j
 				// a user is viewing this path, hint to add it to the cache
 				// (unless we're primarily using binary package cache .a
 				// files).
-				if !h.Config.UseBinaryPkgCache {
+				if !h.Config.UseBinaryPkgCache || req.Method == "textDocument/didSave" {
```
keegancsmith: the main issue with enabling typechecking again, is that it uses a lot of memory. This might be a blocker for this PR / reason to make this opt-in. WDYT @slimsag

lloiser: I have added a simple command line flag `--diagnostics` to activate it

langserver/handler.go
```diff
@@ -397,7 +397,7 @@ func (h *LangHandler) Handle(ctx context.Context, conn jsonrpc2.JSONRPC2, req *j
 				// a user is viewing this path, hint to add it to the cache
 				// (unless we're primarily using binary package cache .a
 				// files).
-				if !h.Config.UseBinaryPkgCache || req.Method == "textDocument/didSave" {
+				if h.Config.DiagnosticsEnabled && (!h.Config.UseBinaryPkgCache || req.Method == "textDocument/didSave") {
```
keegancsmith: I'd suspect this matches the proper intent. We still want to run it if usebinarypkgcache is disabled
```go
!h.Config.UseBinaryPkgCache || (h.Config.DiagnosticsEnabled && req.Method == "textDocument/didSave")
```

keegancsmith:
Sorry one more inline comment then this is ready to merge :)

langserver/handler.go
lloiser: I have added the `Config.DiagnosticsEnabled` check in the `publishDiagnostics`. So there is no harm in doing the typecheck on save too, right?

lloiser: Anyway. I will change it to
```go
if !h.Config.UseBinaryPkgCache || (h.Config.DiagnosticsEnabled && req.Method == "textDocument/didSave") {
```

keegancsmith:
great CI failure is on go tip so is unrelated to your change. Merging in.

keegancsmith:
oh this is merging into `k/better-fmt`! What do you want to do next?

PR #254, opened 2018-03-30
Use ✔ in place of x for supported features

MaskRay:


slimsag:
Nice, thanks! This also better matches the table design on http://langserver.org/

Issue #259, opened 2018-04-21
test case "go install -v all" failures hard to diagnose


seebs:
```
--- FAIL: TestServer (13.80s)
    --- FAIL: TestServer/unexpected_paths (0.78s)
        langserver_test.go:1308: exit status 1
[...]
```

This is sort of opaque.

slimsag:
Fixed via https://github.com/sourcegraph/go-langserver/pull/260

PR #260, opened 2018-04-21
Reorder log/fatal in langserver_test. (Addresses issue #259.)

seebs:
Under circumstances I have not adequately debugged, the
	GOPATH=[tmpdir] go install -v all
used to set up test cases can fail. In my particular case, this appears
to be because something has convinced it that it absolutely has to
rewrite cmd/pprof in $GOROOT.

Reordering these (and adding the GOPATH=...) means that, if the test
fails for this reason, we get a diagnostic more specific than "exit code 1".

Signed-off-by: Seebs <seebs@seebs.net>


slimsag:
LGTM, thanks!

slimsag:
FYI @seebs GitHub is a bit silly and doesn't detect `Addresses issue #259` automatically, you have to remove the `issue` word and use one of these keywords instead for GitHub to close issues on merge: https://help.github.com/articles/closing-issues-using-keywords/

PR #262, opened 2018-04-23
Parallelize findReferencesPkgLevel

josharian:
I'm afraid this diff is hard to read. It is easier to grok if you do a (local) diff ignoring whitespace.


josharian:
> I see in the guru CR there is a semaphore to limit concurrency. I assume this would be useful here as well?

Doh. I meant to ask about that and forgot. :( I was unsure.

I'll add that now, and pull in some more recent changes as well. Stay tuned.

> FYI github has ?w=1 to strip whitespace.

Nice, thanks.

josharian:
Semaphore added, as well as two additional commits.

keegancsmith:
Thanks!

PR #264, opened 2018-04-24
Initial implementation for textdocument/typeDefinition

seebs:
This adds a test case for `textdocument/typeDefinition`, and also an initial implementation.

### DIFF
--- a/langserver/definition.go
+++ b/langserver/definition.go
@@ -7,8 +7,10 @@ import (
 	"go/ast"
 	"go/build"
 	"go/token"
+	"go/types"
 	"log"
 	"path/filepath"
+	"strings"
 
 	"github.com/sourcegraph/go-langserver/langserver/internal/godef"
 	"github.com/sourcegraph/go-langserver/langserver/internal/refs"
@@ -35,6 +37,24 @@ func (h *LangHandler) handleDefinition(ctx context.Context, conn jsonrpc2.JSONRP
 	locs := make([]lsp.Location, 0, len(res))
 	for _, li := range res {
 		locs = append(locs, li.Location)
+	}
+	return locs, nil
+}
+
+func (h *LangHandler) handleTypeDefinition(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) ([]lsp.Location, error) {
+	// note the omission of Godef case; don't want to try to
+	// handle two different ways of doing this just yet.
+
+	res, err := h.handleXDefinition(ctx, conn, req, params)
+	if err != nil {
+		return nil, err
+	}
+	locs := make([]lsp.Location, 0, len(res))
+	for _, li := range res {
+		// not everything we find a definition for also has a type definition
+		if li.TypeLocation.URI != "" {
+			locs = append(locs, li.TypeLocation)
+		}
 	}
 	return locs, nil
 }
@@ -87,6 +107,11 @@ func (h *LangHandler) definitionGodef(ctx context.Context, params lsp.TextDocume
 	return fset, res, []lsp.Location{loc}, nil
 }
 
+type foundNode struct {
+	ident	*ast.Ident   // the lookup in Uses[] or Defs[]
+	typ	types.Object // the type's object
+}
+
 func (h *LangHandler) handleXDefinition(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) ([]symbolLocationInformation, error) {
 	if !util.IsURI(params.TextDocument.URI) {
 		return nil, &jsonrpc2.Error{
@@ -98,7 +123,7 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn jsonrpc2.JSONR
 	rootPath := h.FilePath(h.init.Root())
 	bctx := h.BuildContext(ctx)
 
-	fset, node, pathEnclosingInterval, _, pkg, _, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
+	fset, node, pathEnclosingInterval, prog, pkg, _, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
 	if err != nil {
 		// Invalid nodes means we tried to click on something which is
 		// not an ident (eg comment/string/etc). Return no locations.
@@ -108,14 +133,36 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn jsonrpc2.JSONR
 		return nil, err
 	}
 
-	var nodes []*ast.Ident
+	var nodes []foundNode
 	obj, ok := pkg.Uses[node]
 	if !ok {
 		obj, ok = pkg.Defs[node]
 	}
 	if ok && obj != nil {
 		if p := obj.Pos(); p.IsValid() {
-			nodes = append(nodes, &ast.Ident{NamePos: p, Name: obj.Name()})
+			typ := pkg.TypeOf(node).String()
+			typIdent := typ
+			var typObj types.Object
+			if idx := strings.LastIndex(typ, "."); idx != -1 {
+				typIdent := typ[idx+1:]
+				pkgStr := typ[:idx]
+				typPkg := prog.Package(pkgStr)
+				if typPkg != nil && typPkg.Pkg != nil {
+					scope := typPkg.Pkg.Scope()
+					if scope != nil {
+						typObj = typPkg.Pkg.Scope().Lookup(typIdent)
+					}
+				}
+			} else {
+				for scope := pkg.Pkg.Scope().Innermost(p); typObj == nil && scope != nil && scope != types.Universe; scope = scope.Parent() {
+					typObj = scope.Lookup(typIdent)
+
+				}
+			}
+			nodes = append(nodes, foundNode{
+				ident: &ast.Ident{NamePos: p, Name: obj.Name()},
+				typ: typObj,
+			})
 		} else {
 			// Builtins have an invalid Pos. Just don't emit a definition for
 			// them, for now. It's not that valuable to jump to their def.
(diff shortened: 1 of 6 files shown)


slimsag:
@keegancsmith will probably review this before I can, but I wanted to comment on this part and add some context:

> The implementation does not attempt to handle the Godef case (yet?)

I don't think you should add a godef implementation for this, honestly. The reason `textDocument/definition` has two codepaths (a plain one and a godef one) is because we got a lot of feedback that `textDocument/definition` was slow in comparison to the `godef` CLI so it made sense to very quickly just import that codebase effectively so that the langserver was a competitive alternative to that. I could be wrong -- but I think `godef` doesn't offer this feature (goto type definition) -- so the time would be better spent just making this code path fast :)

slimsag:
Not sure about the fileset stuff off the top of my head.

> I did the other one first because it turns out the default configuration seems to be not to use the Godef stuff.

This is true, the reason for this is that `godef` (and inherently, our godef codepath) relies on the `$GOPATH/pkg` binary files that the Go compiler produces. This is in fact why it is a bit faster / more performant than our other `textDocument/definition` codepath: because it does not need to perform type checking via go/loader.

So, the godef codepath is nice when you're using the langserver in your editor and making small edits constantly and the Go compiler is doing all the heavy lifting of incremental compilation. But it doesn't make sense for us to use that codepath on e.g. sourcegraph.com, where we would need to deal with a whole slew of complexities (persistent disks for `$GOPATH/pkg`, something to prevent race conditions on those files due to parallel requests, etc). Also, from memory, there are some cases where the godef path does not produce as nice results IIRC.

Hopefully if go/types can one day support reusing typechecking information via e.g. an on disk cache, we could eliminate that godef codepath all together.

langserver/definition.go
```diff
@@ -39,6 +41,24 @@ func (h *LangHandler) handleDefinition(ctx context.Context, conn jsonrpc2.JSONRP
 	return locs, nil
 }
 
+func (h *LangHandler) handleTypeDefinition(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) ([]lsp.Location, error) {
+	// note the omission of Godef case; don't want to try to
+	// handle two different ways of doing this just yet.
```
keegancsmith: yeah leave out godef support. We haven't investigated yet, but go's cache dir may enable us to remove godef based lookup.

langserver/definition.go
```diff
@@ -108,14 +133,36 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn jsonrpc2.JSONR
...
-	var nodes []*ast.Ident
+	var nodes []foundNode
 	obj, ok := pkg.Uses[node]
 	if !ok {
 		obj, ok = pkg.Defs[node]
 	}
 	if ok && obj != nil {
 		if p := obj.Pos(); p.IsValid() {
-			nodes = append(nodes, &ast.Ident{NamePos: p, Name: obj.Name()})
+			typ := pkg.TypeOf(node).String()
```
keegancsmith: TypeOf documents that this could return nil. Is it possible in this context? (maybe not due to us doing a type check)

langserver/definition.go
```diff
@@ -108,14 +133,36 @@ func (h *LangHandler) handleXDefinition(ctx context.Context, conn jsonrpc2.JSONR
...
+			typIdent := typ
+			var typObj types.Object
+			if idx := strings.LastIndex(typ, "."); idx != -1 {
```
keegancsmith: I don't think this makes sense for any composite types. See the implementation of how these strings are created and this would likely break structs, maps, etc. https://sourcegraph.com/github.com/golang/go@go1.10/-/blob/src/go/types/typestring.go#L78

You probably need some sort of solution which does the correct thing for each possible type.

seebs:
I am actually not sure whether TypeOf can return nil in this context. Should check for it, anyway.

I think that, if the type isn't named, there's not necessarily a meaningful way to go to the "definition" of the type, so the only non-named type I think we can do anything useful for is *x/**x/etc., where x is a named type. If you have map[string]Foo, it might make sense to guess that you wanted to look at the definition of Foo, but if you have map[Foo]Bar, it's no longer possible to yield a single logically-obvious result, and if you return multiple locations, I don't see a way to label them such that an IDE or something could display a meaningful description of them.

So I think the correct thing is, for pointer types, dereference until we get a non-pointer, and then treat any non-Named type as lacking a point of definition.

seebs:
Okay, I have updated this. It now passes the new tests I added to confirm failures... But it produces a message that appears unrelated to my substantive code changes. Specifically:

2018/04/25 10:54:55 refs.DefInfo: not a package-level definition (ident: x, object: var x []test/pkg/a.A) and unable to follow type (type: []test/pkg/a.A)

I can reproduce this on master by just adding the test case to master, changing wantTypeDefinition to wantDefinition, so I think it's an existing bug/feature/something.

On further study, I've made refs smarter about dereferencing both slices and pointers, and converted both refs and my type lookup to do those recursively. The new test case produces an expected diagnostic from refs for the map case, but I am much less sure what to do about that and would like a second opinion. (My inclination would be to check for expected-failures like that and not emit the diagnostic?)

seebs:
got curious, discovered a cleaner thing:

```
// Dereferencable Represents any go/types.Type which
// has an Elem() interface returning a new Type, which
// includes map, slice, array, and pointer types.
type Dereferencable interface {
	Elem() types.Type
}

func dereferenceType(otyp types.Type) types.Type {
	for {
		switch typ := otyp.(type) {
		case *types.Map:
			return otyp
		case Dereferencable:
			otyp = typ.Elem()
		default:
			return otyp
		}
	}
}
```

I'm not actually sure about the special case for Map, because it turns out that map[string]Foo might well reasonably get a type-lookup for Foo.

langserver/definition.go
```diff
@@ -87,6 +108,26 @@ func (h *LangHandler) definitionGodef(ctx context.Context, params lsp.TextDocume
 	return fset, res, []lsp.Location{loc}, nil
 }
 
+type foundNode struct {
+	ident	*ast.Ident      // the lookup in Uses[] or Defs[]
+	typ	*types.TypeName // the object for a named type, if present
+}
+
+func (h *LangHandler) typeLookup(prog *loader.Program, typ types.Type) *types.TypeName {
```
keegancsmith: nit: make this a free standing function. It doesn't use `h`

langserver/definition.go
```diff
@@ -87,6 +108,26 @@ func (h *LangHandler) definitionGodef(ctx context.Context, params lsp.TextDocume
...
+	if typ == nil {
+		return nil
+	}
+	switch t := typ.(type) {
+	case *types.Pointer:
+		return h.typeLookup(prog, t.Elem())
+	case *types.Slice:
+		return h.typeLookup(prog, t.Elem())
```
keegancsmith: This looks slightly debatable to me. If I look up something which is `[]foo.Bar`, is it confusing to jump to `foo.Bar`? Obviously `foo.Bar` is the more useful thing to look at though.

langserver/internal/refs/refs.go
```diff
@@ -302,7 +302,10 @@ func deepRecvType(sel *types.Selection) types.Type {
 
 func dereferenceType(typ types.Type) types.Type {
 	if typ, ok := typ.(*types.Pointer); ok {
-		return typ.Elem()
+		return dereferenceType(typ.Elem())
+	}
+	if typ, ok := typ.(*types.Slice); ok {
+		return dereferenceType(typ.Elem())
```
keegancsmith: cc @slimsag does this make sense?

seebs:
I've been considering, and I think the dereferenceType() logic might be a better fit for both of the cases in question -- implying it should possibly be a shared function available to both of them, except that I don't know where to put it such that it's reasonable to access it both from internal/refs and the main langserver code. But I think it's a cleaner implementation than the recursive one.

The rationale for dereferencing slices (and also, I think, arrays and chans), is basically that if it makes sense to dereference pointers, it probably also makes sense to dereference any other type which has an unambiguous "element" type on which it's only a qualifier. (Note that a named type which happens to be a slice, pointer, etcetera, won't hit this; it'll be a types.Named with an Underlying() that could be dereferenced.) So since pointers were already being dereferenced, I figured generalizing would be fine.

And yes, typeLookup should probably not be a method. I have no idea why I made it a method.

keegancsmith:
I'm unsure doing the unwrapping for `internal/refs` is what we want though. @slimsag would have to comment.

But I agree with you now, so lets do the unwrapping for everything we can (so channel as well). If we get feedback from users that the behaviour is strange, we can continue the discussion at that point.

seebs:
Thanks for the review comments, I've used the prettier iterative dereferencing logic, and dropped the spurious methodness from typeLookup. (It turns out not to offend me as much to have two implementations, once I realized that typeLookup is actually doing a subtly different thing.)

Still passes tests, still produces a diagnostic from internal/refs on the map test case.

langserver/internal/refs/refs.go
```diff
@@ -300,11 +300,30 @@ func deepRecvType(sel *types.Selection) types.Type {
...
+		switch typ := otyp.(type) {
+			case *types.Map:
+				return otyp
+			case dereferencable:
+				otyp = typ.Elem()
+			default:
+				return otyp
+		}
 	}
-	return typ
```
slimsag: seems like this code is no longer used / was from a prior version of your PR

seebs:
So, looking at the existing wantDefinition testcase:

```"b/b.go:1:72": "/src/test/pkg/b/b.go:1:52-1:53", // declaration of x```

it looks as though the ranges are exclusive on the top end, but inclusive on the bottom end. But... Yes, that means I'm adding an extra +1 that I shouldn't be.

But 1:17-1:18 is right, I think, because it means the range starting at 17, and going up to but not including 18. I think what happened was, I realized that the ranges were non-inclusive, and thought "ahh, that means I need to add one". Fenceposts: My natural enemy.

```node := found.ident```

this is there because there were at least three references to 'node' below, but it doesn't seem like it'd confuse things any to just use found.ident. Also I think leftover from a more complicated previous form before I figured out how the types and ast tree interacted a bit better.

The removal of "return typ" is diff displaying the change confusingly. It helpfully realized that the close-brace right before that line WAS common between both versions of the function, so it carefully split out the removal of "return typ" from the rest of the removal of the previous definition. So the code actually present in the final form is sane:

```
func dereferenceType(otyp types.Type) types.Type {
        for {
                switch typ := otyp.(type) {
                        case *types.Map:
                                return otyp
                        case dereferencable:
                                otyp = typ.Elem()
                        default:
                                return otyp
                }
        }
}
```

it's just showing as two removals because the second-to-last close brace could be a shared line.

So, dropped the extra "node := found.ident", fixed the length calculation and the test, dropped the extra comment. Thanks!

slimsag:
Awesome! That all makes sense

> Fenceposts: My natural enemy.

Mine too 😄

seebs:
The only obvious question would be whether you want the commits squashed or just want to leave the history there. Either's fine by me, I'm past the phase where I need to pretend I don't make mistakes. So I can squash the commits and clean up the commit messages, or just leave as is, whichever you prefer.

slimsag:
I cleaned up (/ removed uncertainty wording) from the PR description + title. I'll just go ahead and squash merge this :)

Issue #280, opened 2018-05-15
Use InitializationOptions instead of flags to drive capabilities


ramya-rao-a:
At present there are 2 flags to the `go-langserver` that control features of the language server
- gocodecompletion
- func-snippet-enabled

One can easily imagine other configuration options we would want to have down the line that controls other features. One such is already being discussed in https://github.com/sourcegraph/go-langserver/issues/272#issuecomment-387347100 which is about the  tool to use for formatting i.e gofmt vs goimports

`InitializationOptions` is something sent to the server from the client which seems like a much better place than flags to control/configure features. Adding more flags to `go-langserver` itself doesnt scale.

cc @keegancsmith @slimsag

ramya-rao-a:
My another concern with flags is that when new flags get added, there is no clean way for the client to know if the user has the version of the language server that supports the said flag. The only way to do this would be to parse the output of `go-langserver -h` and look for all the supported flags

PR #282, opened 2018-05-17
langserver: Support InitializationOptions

keegancsmith:
Fixes https://github.com/sourcegraph/go-langserver/issues/280

cc @lloiser @Contextualist @ramya-rao-a

### DIFF
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -78,8 +78,8 @@ type LangHandler struct {
 	// set by LangHandler creators. Please read config instead.
 	DefaultConfig Config
 
-	// config is the language handler configuration. It is Default with
-	// InitializationOptions applied.
+	// config is the language handler configuration. It is a combination of
+	// DefaultConfig and InitializationOptions.
 	config *Config // pointer so we panic if someone reads before we set it.
 }
 
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -101,6 +101,7 @@ func (h *LangHandler) reset(init *InitializeParams) error {
 			return err
 		}
 	}
+	h.Config = h.Config.Apply(init.InitializationOptions)
 	h.init = init
 	h.cancel = &cancel{}
 	h.resetCaches(false)
--- a/langserver/handler.go
+++ b/langserver/handler.go
@@ -24,9 +24,9 @@ import (
 )
 
 // NewHandler creates a Go language server handler.
-func NewHandler(cfg Config) jsonrpc2.Handler {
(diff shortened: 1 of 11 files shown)


langserver/handler.go
```diff
@@ -73,7 +73,14 @@ type LangHandler struct {
...
 
-	Config Config // language handler configuration; must not change after handling has begun
+	// DefaultConfig is the default values used for configuration. It is
+	// combined with InitializationOptions after initialize. This should be
+	// set by LangHandler creators. Please read config instead.
+	DefaultConfig Config
+
+	// config is the language handler configuration. It is Default with
+	// InitializationOptions applied.
+	config *Config // pointer so we panic if someone reads before we set it.
```
lloiser: config is the language handler configuration. It is a combination of `DefaultConfig` and `InitializationOptions`.

Contextualist:
Is it better if we just take advantage of the standard `ClientCapabilities`?
i.e. 
`config.GocodeCompletionEnabled` -> `init.Capabilities.TextDocument.Completion`
`config.FuncSnippetEnabled` -> `init.Capabilities.TextDocument.Completion.SnippetSupport`

keegancsmith:
> Is it better if we just take advantage of the standard `ClientCapabilities`?
i.e. 
`config.GocodeCompletionEnabled` -> `init.Capabilities.TextDocument.Completion`
`config.FuncSnippetEnabled` -> `init.Capabilities.TextDocument.Completion.SnippetSupport`

cc @ramya-rao-a

ramya-rao-a:
_TLDR_; 

That works for me. It will cause double completions to show up for users who have updated the language server but not the Go extension to VS Code which they can solve by either disabling the language server or updating the Go extension.

_Long version:_

The Go extension for VS Code has no control over the client capabilities being sent to the language server.

`init.Capabilities.TextDocument.Completion` will always be `true` for any VS Code extension that uses the [VS Code Language Client](https://github.com/Microsoft/vscode-languageserver-node/tree/master/client) today. The Client Capabilities only differ from version to version of VS Code client. For example, the `Go to Implementation` feature was a new capability added recently and will not be available in the older versions of the VS Code client.

I'll have to edit the Go extension (and release an update) to not register its own completion provider when the language provider provides one. Users who have updated the language server but not the Go extension, will get double completions.

keegancsmith:
> The Go extension for VS Code has no control over the client capabilities being sent to the language server.

Then lets stick to using the initializationOptions as in this PR. I'll merge this in.

Issue #269, opened 2018-05-02
Hover tooltips for end of line comments


nicksnyder:
It is not uncommon for Go fields to be documented like this (including in the standard library):
```go
type Foo struct {
    Bar string // Bar documentation
}
```

Unfortunately hover tooltips are only provided if comments are structured like this:
```go
type Foo struct {
    // Bar documentation
    Bar string 
}
```

PR #290, opened 2018-06-15
concatenate line comments to field documentation

wbbradley:
This might address issue [269](https://github.com/sourcegraph/go-langserver/issues/269). This was a quick fix, so I'm totally open to feedback on the proper way to go about contributing here. Hope this helps!

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -125,19 +125,14 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 			switch v := path[i].(type) {
 			case *ast.Field:
 				doc = v.Doc
+				// Fallback to inline comments when there is no associated documentation
 				if doc == nil {
 					doc = v.Comment
 				}
 			case *ast.ValueSpec:
 				doc = v.Doc
-				if doc == nil {
-					doc = v.Comment
-				}
 			case *ast.TypeSpec:
 				doc = v.Doc
-				if doc == nil {
-					doc = v.Comment
-				}
 			case *ast.GenDecl:
 				doc = v.Doc
 			case *ast.FuncDecl:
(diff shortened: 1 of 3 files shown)


slimsag:
Hi @wbbradley -- this looks like it would probably be the right approach.

For us to be able to merge this type of change, though, we would need:

1. A test case that fails before and passes after (you could duplicate the `go basic` test case [here](https://github.com/sourcegraph/go-langserver/blob/master/langserver/langserver_test.go#L40-L138) and just keep the `wantHover` section, modifying it to test this functionality)
2. Some inline doc comments before the `if` statements describing why they are there (`// Fallback to inline comments when there is no associated documentation` or something like that).

langserver/hover.go
```diff
@@ -120,25 +120,23 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 
 		// Pull the comment out of the comment map for the file. Do
 		// not search too far away from the current path.
-		var doc *ast.CommentGroup
-		for i := 0; i < 3 && i < len(path) && doc == nil; i++ {
+		var comments string
+		for i := 0; i < 3 && i < len(path) && len(comments) == 0; i++ {
```
keegancsmith: minor nit: `comments == ""`

langserver/util/util.go
```diff
@@ -15,6 +16,19 @@ func trimFilePrefix(s string) string {
 	return strings.TrimPrefix(s, "file://")
 }
 
+// ConcatCommentGroups joins the resultant comment text from two CommentGroups with a newline.
+func ConcatCommentGroups(a, b *ast.CommentGroup) string {
```
keegancsmith: can this just be an unexported helper function in the langserver package rather?

langserver/langserver_test.go
```diff
@@ -912,6 +912,23 @@ type Header struct {
...
+type T struct {
+	Q string // Q is a string field.
+	// X is documented.
+	X int // X has comments.
+}`,
+		},
+		cases: lspTestCases{
+			wantHover: map[string]string{
+				"q.go:3:2": "struct field Q string; Q is a string field. \n\n",
+				"q.go:5:2": "struct field X int; X is documented. \n\nX has comments. \n\n",
```
keegancsmith: should we be stripping whitespace on the right? Either way this looks good and is easy to change :)

PR #295, opened 2018-07-18
set gocode Autobuild to true, allows completion of vendor packages

iamruinous:
Setting autobuild to true will allow gocode to autobuild package cache for vendor packages. I couldn't get dep vendored packages to work without this.


keegancsmith:
Thanks. LGTM. cc @Contextualist

PR #301, opened 2018-08-06
Adds format-tool cli flag

ckaznocha:
Now that multiple format tools are available, I'd like to be able to set one from the CLI!
This PR adds a flag to set the format tool with `goimports` as the default.


keegancsmith:
Thanks!

PR #306, opened 2018-08-14
Run local unit tests inside docker

anjmao:
* Added Makefile with command docker-test to run all tests inside docker.
* Added ./test.sh script which is also checking if files are go formatted.


anjmao:
@keegancsmith Hi, could have a look at this small PR, thanks 👍

keegancsmith:
Thanks. Out of interest why would we want to run the tests in docker? The tests are pretty much self contained and run via `go test`. They should pass on all the platforms.

anjmao:
@keegancsmith Yes, you are right. First I had problem running tests on my osx, I was getting permission denied errors since some tests are writing to file system so then I tried running them inside docker and got it working. And finally I realised I just needed to add `sudo` 😆. I didn't used to use sudo with go test command, but anyway this PR can be closed.

keegancsmith:
I think we write to $TMPDIR (or just /tmp) in our tests. Maybe you have some strange permission issues going on in your OS, like /tmp not being writeable by you?

PR #318, opened 2018-09-12
Add support for linting

### LINKED ISSUE #314: Support alternate linters
Hi!
Some users of my client are looking to use alternate linters for diagnostics, specifically [golangci/golangci-lint](https://github.com/golangci/golangci-lint). Could multiple linters be supported much like multiple formatters are now supported? I'd be happy to send a PR but I may not be able to get to it for a while.

For reference: https://github.com/ckaznocha/ide-go/issues/31

Thanks!

ianlopshire:
This pull requests adds basic support for linting.

- supports the `golint` command
- linting results are published via the diagnostics system
- entire workspace linted on initialization
- current package linted on save

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,16 @@ interface GoInitializationOptions {
    */
   formatTool?: "goimports" | "gofmt";
 
+
+  /**
+   * lintTool decides which tool is used for linting documents. Supported: none and golint
+   *
+   * Diagnostics must be enable for linting to work.
+   *
+   * Defaults to none if not specified.
+   */
+  formatTool?: "none" | "golint";
+
   /**
    * goimportsLocalPrefix sets the local prefix (comma-separated string) that goimports will use.
    *
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ interface GoInitializationOptions {
    *
    * Defaults to none if not specified.
    */
-  formatTool?: "none" | "golint";
+  lintTool?: "none" | "golint";
 
   /**
    * goimportsLocalPrefix sets the local prefix (comma-separated string) that goimports will use.
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ interface GoInitializationOptions {
   /**
    * lintTool decides which tool is used for linting documents. Supported: none and golint
    *
-   * Diagnostics must be enable for linting to work.
+   * Diagnostics must be enabled for linting to work.
    *
    * Defaults to none if not specified.
    */
--- a/langserver/diagnostics.go
+++ b/langserver/diagnostics.go
@@ -43,7 +43,7 @@ func newDiagnosticsCache() *diagnosticsCache {
 
 // publishDiagnostics sends diagnostic information (such as compile
 // errors) to the client.
-func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSONRPC2, diags diagnostics, files []string) error {
+func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSONRPC2, diags diagnostics, source string, files []string) error {
 	if !h.config.DiagnosticsEnabled {
 		return nil
 	}
@@ -53,7 +53,7 @@ func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSON
 	}
 
 	h.diagnosticsCache.update(func(cached diagnostics) diagnostics {
-		return updateCachedDiagnostics(cached, diags, files)
+		return syncCachedDiagnostics(cached, diags, source, files)
 	})
 
 	for filename, diags := range diags {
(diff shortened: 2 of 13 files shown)


ianlopshire:
Can I get a code review on this?

I am still planning on adding support for a couple other linters, but I will open other pull requests for those.

keegancsmith:
Sorry, I've been AFK due to public holidays and conferences on my side. I'll take a look now. In future feel free to at mention myself or slimsag.

keegancsmith:
Apologies, I ran out of time today. It is the top of my TODO list tomorrow to give this a proper review.

langserver/diagnostics.go
```diff
@@ -71,18 +71,31 @@ func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSON
...
+
+		//remove all of the diagnostics for the given source/file combinations and add the new diagnostics to the cache.
+		i := 0
+		for _, diag := range cachedDiagnostics[file] {
+			if diag.Source != source {
+				cachedDiagnostics[file][i] = diag
+				i++
+			}
+		}
+		cachedDiagnostics[file] = append(cachedDiagnostics[file][:i], newDiagnostics[file]...)
```
keegancsmith: If I read the old implementation correctly, `file` is not guaranteed to be in `newDiagnostics`. Otherwise we could just iterator over the `newDiagnostics` map and not pass in the `files` slice.

langserver/diagnostics.go
```diff
@@ -71,18 +71,31 @@ func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSON
...
+
+		// if the file was already in the cache, the existing diagnostics need sent to the client along with the new
+		if fileInCache {
+			newDiagnostics[file] = cachedDiagnostics[file]
+		}
```
keegancsmith: Ok, so if I understand this correctly we need to include diagnostics from other sources for the file. So we can't just rely on what is in newDiagnostics?

langserver/diagnostics.go
```diff
@@ -71,18 +71,31 @@ func (h *LangHandler) publishDiagnostics(ctx context.Context, conn jsonrpc2.JSON
...
+
+		// clear out empty cache
+		if len(cachedDiagnostics[file]) == 0 {
+			delete(cachedDiagnostics, file)
+
+			// clear out the client cache
+			if fileInCache {
 				newDiagnostics[file] = nil
 			}
 		}
```
keegancsmith: so this is unrelated to your code change, but is just so I understand this code again. newDiagnostics is something we send to the client after this function runs. So setting it to nil means we clear it out on the client side.

I wonder if we could simplify all of this and always just send all of cachedDiagnostics (and obviously handle clearing out properly).

langserver/lint.go
```diff
@@ -0,0 +1,199 @@
...
+	err := cmd.Run()
+
+	if err == nil {
+		return nil
+	}
+	if _, ok := err.(*exec.ExitError); ok {
+		return nil
+	}
+
+	return fmt.Errorf("linter check failed: %s (%q)", err, buff.String())
```
keegancsmith: could this function be replaced with a call to https://godoc.org/os/exec#LookPath ?

langserver/lint.go
```diff
@@ -0,0 +1,199 @@
...
+		l, err := buff.ReadString('\n')
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("could not read lint command output: %s", err)
+		}
+
+		l = strings.TrimSuffix(l, "\n")
+		file, line, _, message, err := parseLintResult(l)
```
keegancsmith: you include stderr in buff which probably doesn't conform to the expected input to parseLintResult. parseLintResult also isn't very forgiving. Maybe you shouldn't include stderr and just log it separately?

Additionally there isn't really a reason to read this all into a bytes buffer. Scanner is good at reading an io.Reader line by line.

langserver/diagnostics.go
ianlopshire: That is correct. If there are no issues found in a file, it will not be in `newDiagnostics`. The `files` slice is included for the purpose of clearing out diagnostics that are no longer applicable.

langserver/diagnostics.go
ianlopshire: Notifying the client about diagnostics requires one request per file, so sending all of `cachedDiagnostics` would be a decent amount of extra work.

I do think it could be simplified by having the `diagnostics` that need to be published explicitly returned somewhere instead of `newDiagnostics` being mutated in `syncCachedDiagnostics()`.

I'll spend some time looking at it and seeing if I can clean it up a bit.

README.md
```diff
@@ -49,6 +49,16 @@ interface GoInitializationOptions {
    */
   formatTool?: "goimports" | "gofmt";
 
+
+  /**
+   * lintTool decides which tool is used for linting documents. Supported: none and golint
+   *
+   * Diagnostics must be enable for linting to work.
```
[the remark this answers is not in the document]
ianlopshire: This is fixed.

langserver/handler.go
```diff
@@ -238,6 +242,35 @@ func (h *LangHandler) Handle(ctx context.Context, conn jsonrpc2.JSONRPC2, req *j
...
+			} else {
+				if err := h.linter.IsInstalled(ctx, h.BuildContext(ctx)); err != nil {
+					h.linter = nil
+					log.Printf("warning: lint tool (%s) initialize err: %s", h.config.LintTool, err)
+				}
+			}
+		}
+
+		// kick off a lint of the entire workspace
+		if h.config.DiagnosticsEnabled && h.linter != nil {
```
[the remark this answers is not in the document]
ianlopshire: Yes. This is fixed.

langserver/lint.go
ianlopshire: Yes. This is fixed.

README.md
```diff
@@ -49,6 +49,16 @@ interface GoInitializationOptions {
...
 
+
+  /**
+   * lintTool decides which tool is used for linting documents. Supported: none and golint
+   *
+   * Diagnostics must be enabled for linting to work.
+   *
+   * Defaults to none if not specified.
+   */
+  formatTool?: "none" | "golint";
```
jguenther: `s/formatTool/lintTool/` :)

ianlopshire: Good catch. Fixed!

langserver/diagnostics.go
ianlopshire: I've made some changes that I think help this to be slightly more clear.

langserver/diagnostics.go
keegancsmith: oh yeah I misread this. So if it is missing, it will return a `nil` slice which is safe to pass to the `...` operator. cool.

nicksnyder:
Did this close https://github.com/sourcegraph/go-langserver/issues/314?

PR #329, opened 2018-10-12
ci: Enable windows on travis

keegancsmith:


keegancsmith:
Something strange is happening with the files. I suspect EOL is being changed, which is leading to parsing tests failing w.r.t. hardcoded offsets in the tests. I've tried specifying the relevant gitattribute, but travis seems to be using git archives and there might be some other strange interaction going on.

I'll debug further and likely file a bug report if needs be.

keegancsmith:
Filed an issue here https://travis-ci.community/t/go-cant-find-gcc-with-go1-11-1-on-windows/293

dmitshur:
@keegancsmith What's interesting about the missing `gcc` issue is that it only seems to affect Go 1.11.1 (`1.x`) builds, but not Go 1.9.7 (`1.9.x`) ones. See the 2nd and 3rd bullet points of https://github.com/sourcegraph/go-vcs/pull/109#issuecomment-429500737.

keegancsmith:
@dmitshur we had the same issue with appveyor when go1.11 came out. We had to add mingw to the `%PATH%`. I think its a change in go, since none of our code uses CGO. My best guess is compiling with `-race` in go1.11 triggers compilation of stuff in GOROOT which somehow includes some CGO stuff.

PR #333, opened 2018-10-26
builtin hover docs support

### LINKED ISSUE #174: Hover doesn't work for builtins
I'm not able to get any doc for builtins when using the language server in VS Code.

Before enabling language server, I could hover over a builtin and get doc:
builtin

After enabling language server, nothing happens when I hover over a builtin (`make`, `append`, `len`, `cap`, etc.)

kzh:
screen shot 2018-10-26 at 7 20 23 am

This pull request is to improve documentation preview on hover support for builtin types, functions, and variables. This fixes #174 (Hover doesn't work for builtins).

Todo:
---
1) add unit tests
2) fix incomplete type/function/variable header (append -> `func([]int, int) []int` vs.` func append([]int, int) []int`)
3) heavy refactoring to clean up code

### DIFF
--- a/langserver/hover.go
+++ b/langserver/hover.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sourcegraph/go-langserver/langserver/util"
 	"github.com/sourcegraph/go-lsp"
 	"github.com/sourcegraph/jsonrpc2"
+	"golang.org/x/tools/go/packages"
 )
 
 func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, req *jsonrpc2.Request, params lsp.TextDocumentPositionParams) (*lsp.Hover, error) {
@@ -65,12 +66,18 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 		}
 		return nil, fmt.Errorf("type/object not found at %+v", params.Position)
 	}
-	if o != nil && !o.Pos().IsValid() {
-		// Only builtins have invalid position, and don't have useful info.
-		return nil, nil
-	}
+
 	// Don't package-qualify the string output.
 	qf := func(*types.Package) string { return "" }
+
+	// Handle builtin types, functions and variables.
+	if o != nil && !o.Pos().IsValid() {
+		comments := builtinDoc(node.Name)
+
+		return &lsp.Hover{
+			Contents: maybeAddComments(comments, []lsp.MarkedString{{Language: "go", Value: types.TypeString(t, qf)}}),
+		}, nil
+	}
 
 	var s string
 	var extra string
@@ -92,7 +99,6 @@ func (h *LangHandler) handleHover(ctx context.Context, conn jsonrpc2.JSONRPC2, r
 		if s == "" {
 			s = types.ObjectString(o, qf)
 		}
-
 	} else if t != nil {
 		s = types.TypeString(t, qf)
 	}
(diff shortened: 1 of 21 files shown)


nicksnyder:
Does this fix https://github.com/sourcegraph/go-langserver/issues/174 ?

If so add "fixed https://github.com/sourcegraph/go-langserver/issues/174" to the PR description

langserver/hover.go
```diff
@@ -207,6 +213,83 @@ func packageDoc(files []*ast.File, pkgName string) string {
...
+					return v.Doc
+				}
+			}
+		}
+		return ""
+	}
+
+	if funcDoc := searchFuncs(docs.Funcs); funcDoc != "" {
+		return funcDoc
+	} else if constDoc := searchVars(docs.Consts); constDoc != "" {
```
nicksnyder: Omit unnecessary elses
https://golang.org/doc/effective_go.html#if

```
if a {
    // blah
    return
}
if b {
    // blah
    return
}
```

langserver/hover.go
```diff
@@ -324,8 +407,39 @@ func (h *LangHandler) handleHoverGodef(ctx context.Context, conn jsonrpc2.JSONRP
...
+		// Handle builtin types, functions and variables.
+		if o != nil && !o.Pos().IsValid() {
+			comments := builtinDoc(node.Name)
+
+			return &lsp.Hover{
+				Contents: maybeAddComments(comments, []lsp.MarkedString{{Language: "go", Value: types.TypeString(t, qf)}}),
+			}, nil
+		}
+
 		// TODO: builtins do not have valid URIs or locations.
```
nicksnyder: update this doc to describe how/when this case is reached?

langserver/hover.go
```diff
@@ -324,8 +407,39 @@ func (h *LangHandler) handleHoverGodef(ctx context.Context, conn jsonrpc2.JSONRP
...
 
 	loc := goRangeToLSPLocation(fset, res.Start, res.End)
-
 	if loc.URI == "file://" {
+		_, node, _, _, pkg, _, err := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
+		if err != nil {
+			// Invalid nodes means we tried to click on something which is
+			// not an ident (eg comment/string/etc). Return no information.
+			if _, ok := err.(*invalidNodeError); ok {
+				return nil, nil
```
nicksnyder: what is the difference between returning `nil, nil` here and `&lsp.Hover{}, nil` on L444?

kzh: `&lsp.Hover{}, nil` is returned to indicate that there is a valid identifier being hovered that isn't currently supported (builtins). For invalid identifiers (such as `{`, `=`, `func`, etc), we return `nil, nil`. From the user's perspective, the difference is unnoticeable. 

I've removed the `&lsp.Hover{}, nil` in the next commit.

langserver/hover.go
```diff
@@ -325,9 +433,14 @@ func (h *LangHandler) handleHoverGodef(ctx context.Context, conn jsonrpc2.JSONRP
 
 	loc := goRangeToLSPLocation(fset, res.Start, res.End)
 
+	// Handle builtin objects with invalid locations.
 	if loc.URI == "file://" {
-		// TODO: builtins do not have valid URIs or locations.
-		return &lsp.Hover{}, nil
+		_, node, _, _, _, _, _ := h.typecheck(ctx, conn, params.TextDocument.URI, params.Position)
```
keegancsmith: the only downside here is we try to avoid our typechecker in the `godef` based hover handler since it can be much slower and use more RAM. But given this will only be used on builtins its likely safe to fallback to this.

keegancsmith:
Thanks!

PR #339, opened 2018-11-08
Add support for WebSockets

chrismwendt:
This uses https://github.com/sourcegraph/jsonrpc2 (specifically the [websocket](https://github.com/sourcegraph/jsonrpc2/tree/master/websocket) package) to support WebSocket communication with a client.

I also added log statements whenever a connection is opened/closed tagged with the connection ID (integers starting at 1 for simplicity) to aid debugging:

```
langserver-go: received incoming connection #2
langserver-go: connection #2 closed
```


main.go
```diff
@@ -111,15 +113,48 @@ func run(cfg langserver.Config) error {
...
+			var upgrader = websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
+			connection, err := upgrader.Upgrade(responseWriter, request, nil)
+			if err != nil {
+				log.Println("error upgrading HTTP to WebSocket:", err)
+			}
+			defer connection.Close()
+			connectionCount = connectionCount + 1
+			connectionID := connectionCount
+			log.Printf("langserver-go: received incoming connection #%d\n", connectionID)
+			<-jsonrpc2.NewConn(context.Background(), wsjsonrpc2.NewObjectStream(connection), langserver.NewHandler(cfg), connOpt...).DisconnectNotify()
```
chrismwendt: In contrast to the TCP case, this creates a new handler `langserver.NewHandler(cfg)` for every connection. Without this, the langserver throws an error:

```
language server is already initialized
```

@keegancsmith Do you happen to know if all usages of TCP mode only ever send `initialize` once? I'm guessing lsp-proxy is the only user of TCP mode, and performs connection sharing so that it only sends `initialize` once. Do you foresee any problems with creating a separate handler for each connection? I haven't run across any problems in my experimentation so far, but I just thought I'd check.

main.go
```diff
@@ -111,15 +113,48 @@ func run(cfg langserver.Config) error {
...
+		connectionCount := 0
+
+		mux.HandleFunc("/", func(responseWriter http.ResponseWriter, request *http.Request) {
+			var upgrader = websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
+			connection, err := upgrader.Upgrade(responseWriter, request, nil)
+			if err != nil {
+				log.Println("error upgrading HTTP to WebSocket:", err)
+			}
+			defer connection.Close()
+			connectionCount = connectionCount + 1
```
mbana: Does it sense to execute the rest of this if there was an error in upgrading?

main.go
```diff
@@ -111,15 +113,48 @@ func run(cfg langserver.Config) error {
...
+			}()
 		}
 
+	case "websocket":
+		mux := http.NewServeMux()
+
+		connectionCount := 0
+
+		mux.HandleFunc("/", func(responseWriter http.ResponseWriter, request *http.Request) {
+			var upgrader = websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
```
mbana: I would be tempted to create just one instance of this and reuse it.

main.go
```diff
@@ -111,15 +113,48 @@ func run(cfg langserver.Config) error {
...
+			connection, err := upgrader.Upgrade(responseWriter, request, nil)
+			if err != nil {
+				log.Println("error upgrading HTTP to WebSocket:", err)
```
keegancsmith: you need to return here and likely write a response.

main.go
```diff
@@ -111,15 +113,48 @@ func run(cfg langserver.Config) error {
...
+				log.Printf("langserver-go: disconnected connection #%d\n", connectionID)
+			}()
 		}
 
+	case "websocket":
+		mux := http.NewServeMux()
+
+		connectionCount := 0
+
+		mux.HandleFunc("/", func(responseWriter http.ResponseWriter, request *http.Request) {
```
keegancsmith: in go the idiomatic name for the writer are `w`. For request it is usually `r` or `req`, but `request` isn't that uncommon.

main.go
```diff
@@ -111,15 +113,48 @@ func run(cfg langserver.Config) error {
...
+			defer connection.Close()
+			connectionCount = connectionCount + 1
+			connectionID := connectionCount
+			log.Printf("langserver-go: received incoming connection #%d\n", connectionID)
+			<-jsonrpc2.NewConn(context.Background(), wsjsonrpc2.NewObjectStream(connection), langserver.NewHandler(cfg), connOpt...).DisconnectNotify()
+			log.Printf("langserver-go: disconnected connection #%d\n", connectionID)
+		})
+
+		log.Println("langserver-go: listening for WebSocket connections on", *addr)
+		http.ListenAndServe(*addr, mux)
```
keegancsmith: if this returns an error we probably want to print it and return it. For example if it can't bind to the address, etc it will fail.

main.go
chrismwendt: Got it, fixed by creating a new handler for each connection.

main.go
chrismwendt: I forgot to return early when there's an error upgrading, fixed.

main.go
chrismwendt: Good call, I moved this outside the handler.

main.go
chrismwendt: :+1: Switched to `w`

main.go
chrismwendt: I added a log statement for `http.ListenAndServe`

chrismwendt:
Thanks for the reviews 🙇

PR #345, opened 2018-11-28
Move Go buildserver code from sourcegraph/sourcegraph to go-langserver

chrismwendt:
This moves all of the Go buildserver code from https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/enterprise/cmd/xlang-go into this repo.

The first commit is huge and mechanical (contains all the copied buildserver code as well as `vendor/` changes).

I fixed a few bugs in the integration and proxy tests, but ultimately I had to leave them disabled because they take more than 10 minutes on Windows (Travis's time limit).

TODO

- [x] Determine whether or not `noGoGetDomains` and/or `blacklistGoGet` are still necessary, and if so decide how to plumb them through (command line args, global vars, etc.) **Edit** Yes, these are necessary, and will be read from environment variables
- [x] Decide how to incorporate odds and ends (copy, omit, or factor out?). @keegancsmith I could use your input on this, what do you recommend?
  - [`pkg/tracer`](https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/pkg/tracer) (currently copied)
  - [`pkg/debugserver`](https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/pkg/debugserver) (currently copied)
  - [`pkg/diskcache`](https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/pkg/diskcache) (currently copied)
  - [`pkg/vfsutil`](https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/pkg/vfsutil) (currently copied)
  - [`pkg/gituri`](https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/pkg/gituri) (currently copied)
  - [`pkg/gosrc`](https://github.com/sourcegraph/sourcegraph/tree/fc30b1acc532d0d2ccae95e06017cf28893bb165/pkg/gosrc) (currently copied)
- [x] Get CI to pass


keegancsmith:
copying is fine. Diskcache may make sense as its own package since other projects may benefit from it. (We use it in quite a few places in sourcegraph). But that can be done another time. Ping when CI is happy.

chrismwendt:
@keegancsmith CI is ✅

I also tested this locally (hovers, defs, external refs, etc.) and it seems to work as long as you pass `-usebinarypkgcache=false`. Passing `true` causes errors like "`/src/github.com/gorilla/mux` not found", not sure if it's worth investigating.

keegancsmith:
> I also tested this locally (hovers, defs, external refs, etc.) and it seems to work as long as you pass -usebinarypkgcache=false. Passing true causes errors like "/src/github.com/gorilla/mux not found", not sure if it's worth investigating.

We have binary pkg cache disabled in production @ sourcegraph, so that seems fine given the build server is for that environment. I'll take a look at the PR now.

.travis.yml
```diff
@@ -23,4 +23,4 @@ script:
   - cd $TRAVIS_BUILD_DIR
   # Travis windows is changing line endings, breaking tests. Switch to LF.
   - find langserver -type f -print0 | xargs -0 sed -i ':a;N;$!ba;s/\r//g'
-  - go test -timeout 5m -race ./...
+  - go test -short -timeout 5m -race ./...
```
keegancsmith: so we add this flag to avoid the heavy integration tests in go-buildserver?

buildserver/deps.go
```diff
@@ -0,0 +1,500 @@
...
+// NewDepRepoVFS returns a virtual file system interface for accessing
+// the files in the specified repo at the given commit. The returned VFS
+// is always backed by a zip archive in memory. The following sources are
+// tried in sequence, and the first one that has the repo is used:
+//
+// 1. GitHub's codeload endpoint
+// 2. A full `git clone` followed by `git archive --format=zip <rev>`
+//
+// Source 1 is a performance optimization over cloning the whole repo.
+var NewDepRepoVFS = func(ctx context.Context, cloneURL *url.URL, rev string) (ctxvfs.FileSystem, error) {
```
keegancsmith: what happened to using Sourcegraph's raw API?

> // 1. Directly from gitserver (will be removed in favor of the raw API soon)

buildserver/environment.go
```diff
@@ -0,0 +1,431 @@
...
+		return "", errors.New("unable to determine Go workspace root import path without due to empty root path")
+	}
+	u, err := gituri.Parse(string(originalRootURI))
+	if err != nil {
+		return "", err
+	}
+
+	if path.Join(u.Host, u.Path) == "github.com/golang/go" {
+		return "github.com/golang/go", nil
+	}
```
keegancsmith: why did you add this special case?

buildserver/environment_test.go
```diff
@@ -0,0 +1,168 @@
...
+		{
+			Name:           "monorepo",
+			RootURI:        "git://github.com/alice/monorepo",
+			WantImportPath: "",
+			WantGoPath:     "/workspace/third_party:/workspace/code:/",
+			FS: map[string]string{
+				".vscode/settings.json": fmt.Sprintf(`{
+// this JSON document has comments!
+  "go.gopath": "${workspaceRoot}/third_party%s${workspaceRoot}/code",
+}`, string(os.PathListSeparator)),
```
keegancsmith: nice :)

buildserver/integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+			},
+			wantXDependencies: "gorilla-csrf.json",
+		},
+		"git://github.com/golang/go?f75aafdf56dd90eab75cfeac8cf69358f73ba171": {
+			// SHA is equivalent to go1.7.1 tag, but make sure we
+			// retain the original rev spec in definition results.
+			mode: "go",
+			pinDepReposToRev: map[string]string{
+				"https://github.com/foobar/externalprintf": "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef",
+			},
```
keegancsmith: why was this added?

.travis.yml
chrismwendt: Yup

buildserver/deps.go
chrismwendt: The corresponding lang-go Sourcegraph extension constructs the zip URL to the raw API (looks like https://sourcegraph.com/some/repo/name@somerevision/-/raw/) and passes it to go-langserver. go-langserver doesn't really need to know or care that it points to Sourcegraph's raw API - all that matters from its perspective is that it can HTTP GET a zip archive from the URL.

buildserver/environment.go
chrismwendt: This is a fix for a failing test. I started a code discussion about it a while back, but the link is broken. Here's the conversation:

> @-chrismwendt Why is a canonical import path with the same or fewer `parts` than `popComponents` invalid? I see a test failure when running:
> 
> ```
> $ go test github.com/sourcegraph/sourcegraph/enterprise/cmd/xlang-go/internal/server -run 'TestIntegration'
> --- FAIL: TestIntegration (31.66s)
>     --- FAIL: TestIntegration/golang-go (0.11s)
>         integration_test.go:243: initialize: jsonrpc2: code 0 message: unable to determine workspace's root Go import path: invalid canonical import path "go/constant" in file at path "/src/go/constant/value.go" (original rootPath is "git://github.com/golang/go?f75aafdf56dd90eab75cfeac8cf69358f73ba171") git://github.com/golang/go?f75aafdf56dd90eab75cfeac8cf69358f73ba171
> ```
> 
> Based on https://golang.org/doc/go1.4#canonicalimports, this doesn't look invalid to me https://sourcegraph.sgdev.org/github.com/golang/go@f75aafdf56dd90eab75cfeac8cf69358f73ba171/-/blob/src/go/constant/value.go?diff=f75aafdf56dd90eab75cfeac8cf69358f73ba171&utm_source=chrome-extension#L13
> 
> ```
> package constant // import "go/constant"
> ```
> 
> @-keegancsmith Weird, is this test failing on current master? This is likely a side-effect of not running all tests on CI. This is indeed a correct import path, but only for GOROOT. Is our GOROOT detection not working?
> 
> @-chrismwendt Yes, this is failing on master, but CI skips this test. I filed https://github.com/sourcegraph/sourcegraph/issues/1135 to track this.

buildserver/environment_test.go
chrismwendt: Heh, I found this when the test failed on Travis on Windows.

buildserver/integration_test.go
chrismwendt: Without this, the test code throws an error complaining about an unpinned dependency. As far as I can tell, this repo doesn't actually exist. It's imported by some test data file in go vet https://sourcegraph.com/github.com/golang/go@f75aafd/-/blob/src/cmd/vet/testdata/print.go#L17:21

buildserver/deps.go
chrismwendt: Oh right! I see a couple of ways of solving this:

1. Provide go-langserver with a zip URL template (e.g. `https://sourcegraph.com/%s@%s/-/raw`) somehow (command-line args, env vars, LSP `initializationOptions`, etc.). The raw API would need to support clone URLs in order for this to work, otherwise go-langserver will need to convert clone URLs to repo names via Sourcegraph's GraphQL API, which would go against the goal of removing Sourcegraph-specific code from language servers. @slimsag Any thoughts on this?
2. Add a new LSP request method `resolveZipURL` that the go-langserver can send to the Sourcegraph extension asking for a zip URL provided a clone URL.

Any other ideas?

chrismwendt: I believe a `.netrc` will work for private deps.

I plan to implement option 1 above with the assumption that the repo name is the same as the clone URL, which is usually true. When it's not, the raw API will likely 404 and go-langserver will fall back to a git clone.

chrismwendt: I implemented option 1 (zip template + HEAD request to the raw API, depends on https://github.com/sourcegraph/sourcegraph/pull/1347).

slimsag: > I plan to implement option 1 above with the assumption that the repo name is the same as the clone URL, which is usually true. When it's not, the raw API will likely 404 and go-langserver will fall back to a git clone.

SGTM

chrismwendt: I'll document `.netrc` and SSH keys in https://github.com/sourcegraph/lang-go

PR #348, opened 2018-12-12
Support private dependencies

chrismwendt:
- Adds support for `.netrc` (for username+password basic authentication) on zip URL fetches
- Falls back to git clone when fetching from codeload.github.com 404s so that SSH auth can be attempted

### DIFF
--- a/buildserver/integration_test.go
+++ b/buildserver/integration_test.go
@@ -275,7 +275,7 @@ func useGithubForVFS() func() {
 		if u.Rev() == "" {
 			return nil, errors.Errorf("rev is required in uri: %s", initializeParams.OriginalRootURI)
 		}
-		return vfsutil.NewGitHubRepoVFS(string(u.Repo()), u.Rev())
+		return vfsutil.NewGitHubRepoVFS(context.Background(), string(u.Repo()), u.Rev())
 	}
 
 	return func() {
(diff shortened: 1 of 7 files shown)


buildserver/integration_test.go
```diff
@@ -275,7 +275,7 @@ func useGithubForVFS() func() {
 		if u.Rev() == "" {
 			return nil, errors.Errorf("rev is required in uri: %s", initializeParams.OriginalRootURI)
 		}
-		return vfsutil.NewGitHubRepoVFS(string(u.Repo()), u.Rev())
+		return vfsutil.NewGitHubRepoVFS(context.Background(), string(u.Repo()), u.Rev())
```
keegancsmith:
```suggestion
		return vfsutil.NewGitHubRepoVFS(ctx, string(u.Repo()), u.Rev())
```

vfsutil/zip.go
```diff
@@ -71,3 +97,20 @@ func NewZipVFS(url string, onFetchStart, onFetchFailed func(), evictOnClose bool
...
+	if _, err := os.Stat(netrcFile); os.IsNotExist(err) {
+		return nil
+	}
+	machine, err := netrc.FindMachine(netrcFile, host)
+	if err != nil || machine == nil {
+		return nil
+	}
+	req.Header.Set("Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", machine.Login, machine.Password))))
+	return nil
+}
```
keegancsmith: this never return an error

PR #354, opened 2019-01-14
Write the command line flags to the log

lhchavez:
This change writes the command line flags to the log instead of stdout, since that confuses clients that expect stdout to be a valid JSON-RPC stream.


lhchavez:
This is required to make go-langserver work with [prabirshrestha/vim-lsp](https://github.com/prabirshrestha/vim-lsp) when invoked with nonstandard flags.

PR #360, opened 2019-03-06
Add HTTP basic auth

chrismwendt:
See https://github.com/sourcegraph/sourcegraph/issues/2595


chrismwendt:
Going with a proxy instead https://github.com/sourcegraph/sourcegraph/issues/2595#issuecomment-470014851

PR #363, opened 2019-04-02
Do not include access token in cache key when fetching zip archives

chrismwendt:
Prior to this change, the cache key was the full URL, including whatever username:password was set (e.g. the Sourcegraph access token).

After this change, the cache key will be the URL but with empty username:password.

This will increase the probability of a cache hit because each repo@revision will be fetched once and used by any user.

A `HEAD` request is still made for each user x repo@revision pair to make sure the user has access to the repository contents.

- Helps fix https://sourcegraph.slack.com/archives/C0C324C91/p1554154297030300
- Helps fix https://sourcegraph.slack.com/archives/C0C324C91/p1554159963034900
- Helps `fix` https://github.com/sourcegraph/sourcegraph/issues/1940

cc @beyang just FYI, no action necessary

### DIFF
--- a/vfsutil/zip.go
+++ b/vfsutil/zip.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -22,10 +23,10 @@ import (
 
 // NewZipVFS downloads a zip archive from a URL (or fetches from the local cache
 // on disk) and returns a new VFS backed by that zip archive.
-func NewZipVFS(ctx context.Context, url string, onFetchStart, onFetchFailed func(), evictOnClose bool) (*ArchiveFS, error) {
-	request, err := http.NewRequest("HEAD", url, nil)
+func NewZipVFS(ctx context.Context, urlString string, onFetchStart, onFetchFailed func(), evictOnClose bool) (*ArchiveFS, error) {
+	request, err := http.NewRequest("HEAD", urlString, nil)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to construct a new request with URL %s", url)
+		return nil, errors.Wrapf(err, "failed to construct a new request with URL %s", urlString)
 	}
 	setAuthFromNetrc(request)
 	response, err := ctxhttp.Do(ctx, nil, request)
@@ -33,13 +34,13 @@ func NewZipVFS(ctx context.Context, url string, onFetchStart, onFetchFailed func
 		return nil, err
 	}
 	if response.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("unable to fetch zip from %s (expected HTTP response code 200, but got %d)", url, response.StatusCode)
+		return nil, fmt.Errorf("unable to fetch zip from %s (expected HTTP response code 200, but got %d)", urlString, response.StatusCode)
 	}
 
 	fetch := func(ctx context.Context) (ar *archiveReader, err error) {
 		span, ctx := opentracing.StartSpanFromContext(ctx, "zip Fetch")
 		ext.Component.Set(span, "zipvfs")
-		span.SetTag("url", url)
+		span.SetTag("url", urlString)
 		defer func() {
 			if err != nil {
 				ext.Error.Set(span, true)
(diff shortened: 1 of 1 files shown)


chrismwendt:
@slimsag I added you as a reviewer to help expedite the review of this change since it got bumped in priority today https://sourcegraph.slack.com/archives/C0C324C91/p1554427053097800?thread_ts=1554425326.073300&cid=C0C324C91

vfsutil/zip.go
```diff
@@ -22,24 +23,24 @@ import (
 
 // NewZipVFS downloads a zip archive from a URL (or fetches from the local cache
 // on disk) and returns a new VFS backed by that zip archive.
-func NewZipVFS(ctx context.Context, url string, onFetchStart, onFetchFailed func(), evictOnClose bool) (*ArchiveFS, error) {
-	request, err := http.NewRequest("HEAD", url, nil)
+func NewZipVFS(ctx context.Context, urlString string, onFetchStart, onFetchFailed func(), evictOnClose bool) (*ArchiveFS, error) {
```
slimsag: It's kind of weird to use `urlString` and `urlStruct` naming. I would leave `url` as-is and rename `urlStruct` to just `u`

vfsutil/zip.go
```diff
@@ -54,39 +55,50 @@ func NewZipVFS(ctx context.Context, url string, onFetchStart, onFetchFailed func
...
 			onFetchStart()
-			request, err := http.NewRequest("GET", url, nil)
+			request, err := http.NewRequest("GET", urlString, nil)
 			if err != nil {
-				return nil, errors.Wrapf(err, "failed to construct a new request with URL %s", url)
+				return nil, errors.Wrapf(err, "failed to construct a new request with URL %s", urlString)
 			}
 			request.Header.Add("Accept", "application/zip")
 			setAuthFromNetrc(request)
+			fmt.Println("**** REQ ", urlString)
```
slimsag: remove debug statement

vfsutil/zip.go
```diff
@@ -54,39 +55,50 @@ func NewZipVFS(ctx context.Context, url string, onFetchStart, onFetchFailed func
...
+		// token) so that the same repository at a revision for a different user
+		// results in a cache hit.
+		urlStruct, err := url.Parse(urlString)
+		if err != nil {
+			return nil, err
+		}
+		urlStruct.User = nil
+		urlWithoutUser := urlStruct.String()
+
+		ff, err := cachedFetch(ctx, urlWithoutUser, store, func(ctx context.Context) (io.ReadCloser, error) {
```
slimsag: A cleaner thing to do would just write the auth info stripping as a function:

```Go
func withoutAuthentication(url string) string
```

Then use it here:

slimsag:
```suggestion
		ff, err := cachedFetch(ctx, withoutAuth(url), store, func(ctx context.Context) (io.ReadCloser, error) {
```

vfsutil/zip.go
chrismwendt: What do you suggest for naming now that the the `url` package name conflicts with the local variable `url`?

chrismwendt: Err, hang on, I need to double check the naming here.

slimsag: it won't conflict if you do my `withoutAuth` refactor suggested below

vfsutil/zip.go
slimsag: you are right this func signature would need to rename `url` to `u` though in this case as it would conflict with the import

vfsutil/zip.go
chrismwendt: Like this?

```go
func withoutAuth(urlString string) (string, error) {
	u, err := url.Parse(urlString)
...
```

chrismwendt:
Thanks, @slimsag. I left `urlString` because both `u` and `url` are used for other variables.

felixfbecker:
If the cache key does not include the user/auth info, is there a way as user A I can possibly get data returned that was cached for user B, without being authorised to see that data?

chrismwendt:
> If the cache key does not include the user/auth info, is there a way as user A I can possibly get data returned that was cached for user B, without being authorised to see that data?

go-langserver checks for permission before reading from the cache:

https://github.com/sourcegraph/go-langserver/blob/8e668e29439ada68cd881ec34fc1e11c46129cb6/vfsutil/zip.go#L27-L38

PR #364, opened 2019-04-11
Strip .git suffix from dependencies

chrismwendt:
Some users have import paths that look like `example.com/user/repo.git` and the repository name on Sourcegraph is `example.com/user/repo` (without `.git`).

Prior to this change, go-langserver would attempt to fetch `example.com/user/repo.git` and get a 404.

After this change, go-langserver will fetch `example.com/user/repo` instead. This will break when the repository name on Sourcegraph has a `.git` suffix. @beyang Does that seem acceptable? Alternatively, go-langserver could try the original (with `.git`) and then fallback to dropping the suffix.

This worked in Sourcegraph 2.x probably because the repository contents were obtained through lsp-proxy and gitserver, rather than the raw API.


beyang:
Trying and then falling back seems better than stripping the `.git` suffix altogether (as the latter might cause further unanticipated failures).

chrismwendt:
Switched to trying then falling back.

PR #369, opened 2019-05-09
LSP-conformant originalRootURI

chrismwendt:
**Background**

Because `originalRootURI` is not part of the [LSP specification for `InitializeParams`](https://microsoft.github.io/language-server-protocol/specification#initialize), clients cannot use LSP-conformant libraries such as https://github.com/sourcegraph/lsp-client without adding a workaround specifically for communicating with go-langserver.

There is an untyped field, `InitializeParams.initializationOptions`, which can easily be used to send the `originalRootURI`.

**Details**

After this change, go-langserver will read the `originalRootURI` from `InitializeParams.initializationOptions` if it's absent from the top level `InitializeParams`. It will still be compatible with existing clients that currently put `originalRootURI` in the top level `InitializeParams`.

**One other alternative considered**

Alternatively, the `InitializeParams.rootURI` might be able to carry the current value of `originalRootURI` because it's currently set to [the informationless value of `file:///` in sourcegraph-go](https://github.com/sourcegraph/sourcegraph-go/blob/4ee437f3d2526e1c63e511be048a840e320e013b/src/lang-go.ts#L236). However, this change [seems difficult](https://docs.google.com/a/sourcegraph.com/document/d/1rFcwzKaEf2WtMJBZAc4L1LxBJqgKAduI8PMvjnNowE0/edit?disco=AAAAC_qY_8c) to make in go-langserver:

> After attempting to move originalRootURI to rootUri for ~2 hours, I'm finding that the assumption that rootUri is `file:///` pervades the LSP request handlers, utility functions, and glue code. Changing it will take longer than I had expected and I'm concerned it might introduce bugs despite the pretty good test coverage of go-langserver.

This helps integrate lsp-client into the Go extension https://github.com/sourcegraph/sourcegraph-go/issues/28

### DIFF
--- a/buildserver/build_server.go
+++ b/buildserver/build_server.go
@@ -178,6 +178,19 @@ func (h *BuildHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jso
 			return nil, err
 		}
 
+		// Fall back to reading the `originalRootURI` from
+		// `InitializeParams.intitializationOptions` if absent from the
+		// `InitializeParams`. This makes go-langserver more LSP-compliant so that
+		// clients don't need to go outside of the LSP specification to communicate
+		// with it.
+		if params.OriginalRootURI == "" {
+			if initializationOptions, ok := params.InitializationOptions.(map[string]interface{}); ok {
+				if originalRootURI, ok := initializationOptions["originalRootURI"].(string); ok {
+					params.OriginalRootURI = lsp.DocumentURI(originalRootURI)
+				}
+			}
+		}
+
 		if Debug {
 			var b []byte
 			if req.Params != nil {
--- a/buildserver/proxy_test.go
+++ b/buildserver/proxy_test.go
@@ -22,24 +22,33 @@ import (
 	"github.com/sourcegraph/jsonrpc2"
 )
 
+type OriginalRootURIParent int
+
+const (
+	parentInitializeParams OriginalRootURIParent = iota
+	parentInitializationOptions
+)
+
 func TestProxy(t *testing.T) {
 	if testing.Short() {
 		t.Skip()
 	}
 
 	tests := map[string]struct {
-		rootURI           lsp.DocumentURI
-		mode              string
-		fs                map[string]string
-		wantHover         map[string]string
-		wantDefinition    map[string]string
-		wantXDefinition   map[string]string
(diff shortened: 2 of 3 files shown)


buildserver/build_server.go
```diff
@@ -178,6 +178,19 @@ func (h *BuildHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jso
 			return nil, err
 		}
 
+		// Fall back to reading the `originalRootURI` from
+		// `InitializeParams.intitializationOptions` if absent from the
+		// `InitializeParams`. This makes go-langserver more LSP-compliant so that
+		// clients don't need to go outside of the LSP specification to communicate
+		// with it.
```
felixfbecker: Note that this is still not really compliant from a semantics perspective because `initializationOptions` is intended to be options you would pass through CLI arguments, i.e. not values that change for every workspace.

felixfbecker:
> After attempting to move originalRootURI to rootUri for ~2 hours, I'm finding that the assumption that rootUri is file:/// pervades the LSP request handlers, utility functions, and glue code. Changing it will take longer than I had expected and I'm concerned it might introduce bugs despite the pretty good test coverage of go-langserver.

Could you give some examples of such assumptions?

chrismwendt:
Here are a few examples:

https://github.com/sourcegraph/go-langserver/blob/33d2968a49e1131825a0accdd64bdf9901cf144c/buildserver/environment.go#L104

https://github.com/sourcegraph/go-langserver/blob/33d2968a49e1131825a0accdd64bdf9901cf144c/langserver/handler_common.go#L32-L35

https://github.com/sourcegraph/go-langserver/blob/33d2968a49e1131825a0accdd64bdf9901cf144c/langserver/loader.go#L40

Unless this PR looks ok, I'll try moving forward with moving `originalRootURI` to `rootUri` tomorrow. Currently, same-repo j2d returns `file:` URIs whereas cross-repo j2d returns `git:` URIs, and I think that moving `originalRootURI` to `rootUri` will make that uniform so they're all `git:` URIs, which would simplify the Go Sourcegraph extension a bit and avoid the need for another parameter to the `serverToClientURI` function in lsp-client.

chrismwendt:
Closing in favor of https://github.com/sourcegraph/go-langserver/pull/373

PR #370, opened 2019-05-09
Derive zip URL from template

chrismwendt:
Passing both `zipURL` and `zipURLTemplate` is redundant. `zipURL` can be derived by instantiating the `zipURLTemplate` with `originalRootURI`.

This adds support for clients that omit `zipURL` and only provide `zipURLTemplate`. This is part of [simplifying lsp-client's communication with go-langserver](https://docs.google.com/document/d/1rFcwzKaEf2WtMJBZAc4L1LxBJqgKAduI8PMvjnNowE0/edit#heading=h.44mn607i8nf4).

### DIFF
--- a/buildserver/vfs.go
+++ b/buildserver/vfs.go
@@ -34,6 +34,9 @@ var RemoteFS = func(ctx context.Context, initializeParams lspext.InitializeParam
 			return zipURL
 		}
 		zipURLTemplate, _ := initializationOptions["zipURLTemplate"].(string)
+		if zipURLTemplate == "" {
+			return ""
+		}
 		root, err := url.Parse(string(initializeParams.OriginalRootURI))
 		if err != nil {
 			return ""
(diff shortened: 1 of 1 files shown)


buildserver/vfs.go
```diff
@@ -26,8 +29,16 @@ var RemoteFS = func(ctx context.Context, initializeParams lspext.InitializeParam
...
+		zipURL, _ := initializationOptions["zipURL"].(string)
+		if zipURL != "" {
+			return zipURL
+		}
+		zipURLTemplate, _ := initializationOptions["zipURLTemplate"].(string)
+		root, err := url.Parse(string(initializeParams.OriginalRootURI))
+		if err != nil {
+			return ""
+		}
+		return fmt.Sprintf(zipURLTemplate, path.Join(root.Host, root.Path), root.RawQuery)
```
keegancsmith: you may want to test this with an empty zipURLTemplate.

chrismwendt: Ah, `fmt.Sprintf("", ...)` results in something like `%!(EXTRA string=foo)`. I added a check for this.

PR #373, opened 2019-05-14
Support git:// in the rootUri

chrismwendt:
**This change only affects the buildserver, which is only used by Sourcegraph.** There is no change to editor users.

**Motivation**

I'm working on [integrating lsp-client into the Go extension](sourcegraph/sourcegraph-go#28), which has surfaced a non-spec-compliant characteristic of go-langserver:

- `originalRootUri` must be specified in the `InitializeParams`

Instead of adding workarounds in lsp-client, I'm adding support in go-langserver for putting the value of `originalRootUri` in the `rootUri` field and omitting `originalRootUri`.

**Backwards compatibility**

During a migration period in which customers bump their go-langserver versions, there will be some customers running the old version and some running the new version.

My current plan is to only merge lsp-client into the Go extension when customers have bumped their versions. This means that in the meantime, go-langserver must remain compatible with the current Go extension (i.e. it must still support `originalRootUri`). That's why you'll see the flag `clientUsesFileSchemeWithinWorkspace` and some `if`-statements in this PR.

**Details**

In the `rootUri` in the `InitializeParams` of the `initialize` request, clients can send either:

- (not new in this PR) A `file://` URI, which indicates that:
  - Same-workspace file paths will also be `file://` URIs
  - Out-of-workspace file paths will be `git://` URIs
  - `originalRootUri` is present
- (new in this PR) A `git://` URI, which indicates that:
  - Both same-workspace and out-of-workspace file paths will be non-`file://` URIs
  - `originalRootUri` is absent and `rootUri` contains the original root URI

This subsumes the other approach https://github.com/sourcegraph/go-langserver/pull/369


buildserver/build_server.go
```diff
@@ -327,10 +349,27 @@ func (h *BuildHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jso
...
+				if !strings.HasPrefix(string(uri), "file:///") {
+					return uri // refers to a resource outside of this workspace
+				}
+				path = strings.TrimPrefix(string(uri), "file://")
+			} else {
+				currentURL, err := url.Parse(string(uri))
+				if err != nil {
+					return uri
+				}
+				originalRootURI, err := url.Parse(string(h.init.OriginalRootURI))
```
keegancsmith: minor: Anyway we can avoid parsing this here? This function can be called a lot, especially for large results.

buildserver/proxy_test.go
```diff
@@ -42,8 +44,9 @@ func TestProxy(t *testing.T) {
 		depFS             map[string]map[string]string // dep clone URL -> map VFS
 	}{
 		"go basic": {
-			rootURI: "git://test/pkg?deadbeefdeadbeefdeadbeefdeadbeefdeadbeef",
-			mode:    "go",
+			rootURI:         "file:///",
+			originalRootURI: "git://test/pkg?deadbeefdeadbeefdeadbeefdeadbeefdeadbeef",
```
keegancsmith: given we are supporting both use cases while we roll out, can't you adjust each test case to run twice. Once with the old rootURI, and one with the new way of specifying?

buildserver/build_server.go
```diff
@@ -327,10 +349,27 @@ func (h *BuildHandler) Handle(ctx context.Context, conn *jsonrpc2.Conn, req *jso
...
+				if err != nil {
+					return uri
+				}
+				if currentURL.Hostname() == originalRootURI.Hostname() && currentURL.RawPath == originalRootURI.RawPath {
+					path = currentURL.Fragment
```
keegancsmith: shouldn't you be checking everything but fragment. IE so need to also include protocol. Maybe you could do something like

```go
path = currentURL.Fragment
currentURL.Fragment = ""
if *currentURL != *originalRootURI {
  return uri
}
```

Alternatively, why can't we just do `if strings.HasPrefix(uri, h.init.OriginalRootURI)`?

buildserver/build_server.go
chrismwendt: ✅ [Do not parse OriginalRootURI so often](https://github.com/sourcegraph/go-langserver/pull/373/commits/f1f8c3d5674c2aa4b2fe5ca267fbd5b1d74be42a)

buildserver/proxy_test.go
chrismwendt: ✅ [Test both with and without OriginalRootURI](https://github.com/sourcegraph/go-langserver/pull/373/commits/86afb912a487cc077cfbbb55bdca5667914e37b9)

buildserver/proxy_test.go
```diff
@@ -131,92 +131,6 @@ func TestProxy(t *testing.T) {
 				"a_test.go:1:46": "A int",
 			},
 		},
-		"go subdirectory in repo": {
```
chrismwendt: This removes a test for what seems to be a currently unused feature: where the root URI is set to a subdirectory in a repository. The Go extension always sets the root URI to the top level of the repository. Does anyone know of a use case for this? Optimization? Perhaps this can be brought back when the Go extension starts making use of it.

buildserver/build_server.go
chrismwendt: ✅ [Check everything but the fragment of the URL](https://github.com/sourcegraph/go-langserver/pull/373/commits/1c0778e6ba5a4db7af7d5b572218224828f7e42b)

buildserver/proxy_test.go
keegancsmith: this was part of the original design by @sqs. I don't think we ever used this feature.

felixfbecker: That sounds like a really good feature to have though that can be supported with the new root URIs

chrismwendt: In the code nav sync today, we decided that we don't need to support this right now. We can bring this back later without incurring tech debt in the meantime.

chrismwendt:
Thanks for the review 🙇

PR #382, opened 2020-03-30
Use canonical import for log15

mrnugget:
This changes the import path for `log15`, inspired by [this PR](https://github.com/sourcegraph/sourcegraph/pull/9245).

### DIFF
--- a/langserver/internal/refs/refs_test.go
+++ b/langserver/internal/refs/refs_test.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


langserver/internal/refs/refs_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"strings"
 	"testing"
 
-	_ "gopkg.in/inconshreveable/log15.v2"
+	_ "github.com/inconshreveable/log15"
```
creachadair: The `init` for log15 doesn't appear to do anything relevant—is this import even needed?

